PR #567, opened 2015-01-27
Added option to write junit file to specified directory instead of console

thebostik:
This was present in 2.5.7 and I still have jenkins set up to grab the junit file from a directory.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -225,7 +225,20 @@ var fs = require('fs-extra'),
     },
     junit: {
       flag: true,
-      help: 'Create JUnit output to the console.'
+      help: 'Create JUnit output. By default, prints to the console.'
+    },
+	junitDir: {
+      metavar: '<DIR>',
+      help: 'JUnit output xml file will be written to specified directory instead of console. Use with --junit option.',
+      callback: function(file) {
+        if (!fs.existsSync(file)) {
+          fs.mkdirs(file, function(err) {
+            if (err) {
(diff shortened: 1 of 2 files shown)


soulgalore:
Hey Stephen, isn't possible to just pipe it to which file you want? When I changed it for 3.0 I thought it was cleaner to put it to the console and then yourself could choose to pipe it to a file. Or do you have a specific problem where it don't work?

Best
Peter

thebostik:
Yes, that's what I did at very first, but noticed at least one occasion other stuff got output, perhaps stderr? I guess as an option I would just do it this way to guarantee it's a valid junit format, though for others needs stdout may work better and didn't want to change your default.

soulgalore:
hmm then there's a defect, all errors should only be sent to the log when you run it like that. do you still have it or can you reproduce it? would like to fox the original problem first :)

thebostik:
Yes, after further consideration and a bit of trying to replicate, it was probably due either to browsertime hanging when trying to use Chrome or troubleshooting phantom2 even I may have just written some stuff to stdout. I switched to a hardcoded file path for junit output almost immediately without too much thought.

I think I could live with redirected stdout junit, just wasn't sure if other stuff might get garbled in.

soulgalore:
ok, lets leave it for now, and then I'll check it later.

soulgalore:
Closing this since we have rewritten everything in 4.0. Please create a new PR for changes in 4.0.

PR #575, opened 2015-02-08
Fix for exception throwing Java version check

stefanjudis:
Just run in this issue.

At my machine:

```
stefan @ stefan-mac: ~/Desktop/sitespeed-test
> java -version                                          [12:50:24]
openjdk version "1.7.0-jdk7u4-b21"
OpenJDK Runtime Environment (build 1.7.0-jdk7u4-b21-20120427)
OpenJDK 64-Bit Server VM (build 23.0-b21, mixed mode)
```

So `matches` becomes `null` and sitespeed is crashing. Fixed it in this PR.

### DIFF
--- a/lib/util/util.js
+++ b/lib/util/util.js
@@ -358,8 +358,8 @@ module.exports = {
             if (err) {
               return callback(err);
             }
-            var matches = stderr.match(/java version "(.*)"/);
-            var version = matches.length === 2 ? matches[1] : 'unknown';
+            var matches = stderr.match(/(java version|openjdk version) "(.*)"/);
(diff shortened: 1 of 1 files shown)


tobli:
Hi Stefan,
Thanks for the patch! I didn't test with an OpenJDK installation before. I'd prefer a solution that matches either java version or openjdk version. If you'd like to rework the solution to do that it'd be great.
/Tobias

stefanjudis:
@tobli sure thing. give me a few min. :)

stefanjudis:
@tobli done. :)

tobli:
Great work, thanks!

PR #675, opened 2015-05-27
Ability to avoid sandbox mode when running Chrome.

yesman82:
This may resolve issues when running from within continuous integration systems (e.g. Jenkins) or IDEs.

See: https://sites.google.com/a/chromium.org/chromedriver/help/chrome-doesn-t-start

This is a follow-up commit to support the new browsertime feature (https://github.com/tobli/browsertime/pull/85)


yesman82:
Missed merging latest head :(

PR #676, opened 2015-05-27
Ability to avoid sandbox mode when running Chrome.

yesman82:
This may resolve issues when running from within continuous integration systems (e.g. Jenkins) or IDEs.

See: https://sites.google.com/a/chromium.org/chromedriver/help/chrome-doesn-t-start

This is a follow-up request for browsertime pull request https://github.com/tobli/browsertime/pull/85


soulgalore:
So I can just close this right? It's implemented in Browsertime and we can pass it using btConfig.

yesman82:
correct

PR #679, opened 2015-06-05
enhanced the getGraphiteURLKey function to replace additional characters

### LINKED ISSUE #651: Change key structure of the page URL sent to Graphite
> To make those calculations easier per-site-per-page I would suggest I slightly different metric naming send to Graphite for "per page" metrics: I would suggest the following segments: protocol.hostname.pathname (changing dots to underscore for hostname). This way all urls would have the same four metric pieces. Would make templating in Grafana allot easier as well.
> @JeroenVdb

Yes good idea, lets fix this!

EikeDawid:
Fixes #651 

I run into some issues with certain characters on URI that where clashing with graphite functionality or caused parsing issues for grafana when trying to retrieve data from graphite

The characters are:

* pipe (|)
* comma (,)
* plus (+)

... they are replaced by a ```_``` (as the others are)

### DIFF
--- a/lib/util/util.js
+++ b/lib/util/util.js
@@ -300,7 +300,7 @@ module.exports = {
         pathName = 'slash';
       }
 
-      var replace = ['.', '~', ' ', '/'];
+      var replace = ['.', '~', ' ', '/', '+', '%7C', ','];
       replace.forEach(function(replaceMe) {
         if (pathName.indexOf(replaceMe) > -1) {
           pathName = pathName.split(replaceMe).join(char);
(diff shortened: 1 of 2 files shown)


soulgalore:
Hello @EikeDawid , I've changed so we use pipe char instead of %7C , I think the URL should already be decoded or have you seen that it don't work in your case?

Best
Peter

EikeDawid:
@soulgalore: I had to do it cause ```url.parse(theUrl);``` returns the ```|``` encoded on my node version. 

However i just checked the build (on travis) and i see the encoded version is failing on node 0.10, while the un-encoded fails on 0.12. 

So that means that ```url.parse(theUrl);``` behaves differently on both. 

That's a bit of a bummer - there are some other changes too:
https://github.com/joyent/node/issues/8332

Sorry, that i did not check on an older node version ....

I think the fastest way to make it work consistently would be to add both versions (encode and unencoded).  Shall i send another pull request with that?

soulgalore:
@EikeDawid I'll fix it now, that's no problem. When I checked your test, I didn't see you added two different tests for pipe, that's nice. I'll check it in now and see if it works :)

PR #686, opened 2015-06-11
removed line-breaks from travis yaml.

cgoldberg:
there were some spurious line-breaks in the middle of the .travis.yml config file.   This nitpick PR removes them :)

(It seems to get parsed ok, but it's confusing to read for humans).


soulgalore:
thanks @cgoldberg !

PR #719, opened 2015-08-08
use query string in url for graphite

jeremy-green:
We use query string to toggle things like ads and Optimizely.  The tests are run against the same URL's with ads on and off.  I noticed our scores seemed a little too forgiving for a site with ads.  It looks like query strings aren't taken into account when sending the data to Graphite.

This change combines the query string to an object which then converts it to an underscore delimited string.  It takes into account an empty value and filters empty strings when returning the `getGraphiteURLKey`.

### DIFF
--- a/lib/util/util.js
+++ b/lib/util/util.js
@@ -291,7 +291,7 @@ module.exports = {
       var join = '.';
       var char = '_';
 
-      var myUrl = url.parse(theUrl);
+      var myUrl = url.parse(theUrl, true);
       var protocol = myUrl.protocol.replace(':', '');
       var hostname = myUrl.hostname.split('.').join(char);
       var pathName = myUrl.pathname;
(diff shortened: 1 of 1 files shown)


soulgalore:
Hey @jeremy-green ,
ah great thanks for the PR. Yep, I've missed that! I'll do a small fix release on Sunday and then merge this and do a release sometimes next week, need to be bumped since it will change behavoir (but for the good).

Will check the implementation then.

Thanks again for the PR!!!

soulgalore:
sorry @jeremy-green for not getting back on this. I'll merge it and release it this weekend, will add a flag though so until 4.0 you need to add an extra flag to get the parameter, just to not break for other users, ok?

jeremy-green:
That sounds great.  Thanks!!

soulgalore:
Ah I forgot to ping you @JeroenVdb about this (I thought I had but must have forgotten). Thinking about how to solve this the best way. Now the we add a new path segment for the query like path.query but I think we should keep number of paths so we (if we have a query string) add it to the path: pathquery. Or what do you think is the best?

P.S will add this behind a flag.

PR #720, opened 2015-08-09
Inline support for requestHeader config Fixes #715

### LINKED ISSUE #715: inline support for requestHeader config
Right now --requestHeaders config needs a FILE to set value:
--requestHeaders FILE
For example:
sitespeed.io -u http://www.example.com --requestHeaders cookie.json

I'd like to set the value as inline like this:
sitespeed.io -u http://www.example.com --requestHeaders "{'Cookie':'12345'}"

This should be fairly simple improvement by checking the value whether it has .json at the end or a valid json string etc...

tufandevrim:


soulgalore:
Many thanks @tufandevrim ! I'll merge tomorrow Monday and do a release!

Issue #691, opened 2015-07-01
The screenshots are not generated


shaky:
Hi guys,
First of all I want to thank you for all the hard work. The tool is amazing!
Looks like the screenshots are not generated or my configuration is wrong.

Here's the command I'm running:
sitespeed.io -u URL -b chrome --profile mobile --screenshot --connection mobile3g -d 2

This is the info line:
info: OS: 'darwin 13.4.0', Node.js: 'v0.10.29', sitespeed.io: '3.6.3', PhantomJS: '1.9.8', java: '1.7.0_65', browsertime: '0.9.8'

Thanks for help!

soulgalore:
Hi @shaky ,
ok, can you send me the logs from the run (in your result folder there's a sitespeed.io.log, browsertime.log and browsermobproxy.log) then I can see what's happening. 

My email: peter [AT] soulgalore [dot] com

Thanks for reporting! 
Peter

shaky:
Hi @soulgalore 
Thanks for reply.
I have just emailed you the logs.
Cheers!
Raz

jzoldak:
Hi @soulgalore - first echo'ing @shaky's thanks for your awesome project.
FWIW I'm also am unable to capture screenshots. 
Here's the relevant section from my sitespeed.io.log.
If I get some time tomorrow I'll try to dig into it a bit.
```
2015-08-11T19:11:13.395Z - info: [sitespeed] OS: 'linux 3.2.0-65-virtual', Node.js: 'v0.10.37', sitespeed.io: '3.8.0', PhantomJS: '1.9.8', java: '1.7.0_79', browsertime: '0.10.2'
[...]

2015-08-11T19:12:51.378Z - info: [sitespeed] Taking screenshot for <URL for my page under test>
2015-08-11T19:13:03.453Z - error: [sitespeed] Phantomjs output for '<URL for my page under test>' (stdout): SyntaxError: Unable to parse JSON string

  /usr/lib/node_modules/sitespeed.io/lib/headless/scripts/screenshot.js:32

2015-08-11T19:13:03.456Z - error: [sitespeed] Error taking screenshots for url '<URL for my page under test>': { [Error: Command failed: ] killed: true, code: null, signal: 'SIGTERM' }
```

soulgalore:
Hi @jzoldak ,
thanks again :)

[...]

PR #724, opened 2015-08-11
All args to the screenshot script should be passed in the right order

jzoldak:
@soulgalore I think that this should fix my screenshot problem, which I posted under #691.

I can't test it out right now though because my target environment is a bit messed up at the moment. I'll do that tomorrow.


soulgalore:
thanks @jzoldak

jzoldak:
@soulgalore I tested this out and it works for me now. \o/
Thanks very much for getting this in so quickly.
Sorry about the missing curly brace. :|

soulgalore:
thank you @jzoldak for fixing the PR :)

I'll be travelling the coming days but think this will be released on Sunday, will try to squeeze in a couple of more fixes.

PR #727, opened 2015-08-17
All args to the headless script should be passed in the right order

jzoldak:
Similar issue to #724.
This one fixes the usage of " --browser headless" with basic auth and request headers.


soulgalore:
Hey @jzoldak ,
thanks, will merge it tomorrow. Do you need the fix asap or is it more that you found the defect?

Best
Peter

jzoldak:
Super, thanks @soulgalore. 
No huge rush - I am using the feature, but I can also just install off my fork in the interim.

soulgalore:
Cool, I'll probably do a release this weekend then, trying it in some more fixes. Thanks again for the PR.

PR #728, opened 2015-08-18
#651 Create same Graphite namespace structure for the domain url as for per-page metrics

JeroenVdb:
This should fix issue #651.

Crawling **http://www.jeroenvdb.be** creates
````
sitespeed.io.http.www_jeroenvdb_be.slash.*
sitespeed.io.summary.http.www_jeroenvdb_be.slash.*
````
instead of www_jeroenvdb_be and www.jeroenvdb.be:
````
sitespeed.io.http.www_jeroenvdb_be.slash.*
sitespeed.io.summary.www.jeroenvdb.be.*
````

### DIFF
--- a/conf/defaultConfig.js
+++ b/conf/defaultConfig.js
@@ -13,6 +13,7 @@ var defaultConfig = {
   graphitePort: 2003,
   graphiteNamespace: 'sitespeed.io',
   graphiteData: 'all',
+  updatedGraphiteDomainSection: false,
   resultBaseDir: 'sitespeed-result',
   viewPort: '1280x800',
   waitScript: ' if (window.performance && window.performance.timing)'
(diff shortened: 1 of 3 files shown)


JeroenVdb:
To discuss: do we leave the trailing .slash.* for the summary namespace or not... It won't hurt to leave it I guess.

soulgalore:
Hi @JeroenVdb ,
thanks for fixing this, have been focusing on so many other things the last weeks. Do you also have time to add a flag, so we can choose from the CLI so we don't break the current impl and then change it for real in the next major release?

Think we should remove the slash (didn't even remember that it is there).

Thanks again 
Peter

JeroenVdb:
@soulgalore any idea for a name for that flag?
````
--updatedGraphiteNamespace=true ?
````

soulgalore:
names are always the problem :) Adding <pre>--updatedGraphiteDomainSection</pre> should do it if @tobli doesn't have a better suggestion?

JeroenVdb:
@soulgalore added the --updatedGraphiteDomainSection flag for this change.

Tested with --file, --sites and --url, with and without flag.

soulgalore:
Cool, merge this tonight and release it this weekend, many thanks @JeroenVdb !!!

soulgalore:
thanks again @JeroenVdb !

PR #746, opened 2015-09-16
Update testRenderer.js

laer:
Fixing typo, because of it grunt-sitespeedio fails the build no matter if GPSI score matches the budget


soulgalore:
ah thanks @laer will have a look tonight.

soulgalore:
thank you @laer !!!

Issue #751, opened 2015-09-23
Add ability to specify per-rule budgets rather than a single default


jzoldak:
I can't figure out how to configure budgets on a per-rule basis. All examples I found only set a value for 'default' under 'rules'.
[The code](../blob/master/lib/tests/testRenderer.js#L288) seems to back this up.

I want to be able to set the values for each rule, acknowledging that the current application has issues that we are working on and preventing regression from this point. Something like this:
```
{
  "rules": {
    "default": 90,
    "criticalpath": 10,
    "jsnumreq": 85,
    "yminify": 70,
    "ycookiefree": 25,
    "syncjsinhead": 60,
    "expiresmod": 75,
    "longexpirehead": 70
  }
}
```
Or maybe I may be misunderstanding and it is possible currently?

soulgalore:
Hmm, I need to check it out, we use to have it but I think it actually has been removed during the ride. Bit we should aim for it in 4.0. What you can do is send a list of rules that you don't want to include, I mean that's not perfect but maybe it could help a bit.

Best
Peter

soulgalore:
thanks again!

PR #752, opened 2015-09-23
Add ability to budget on a per rule basis

jzoldak:
Fix for #751.
@soulgalore I tested this locally.


soulgalore:
thanks @jzoldak !

Issue #791, opened 2015-12-15
when running sitespeed.io through a proxy the screenshot is not taken properly


daehin:
I am testing a proxy server for ad blocking. when i run sitespeed.io with the -p option it runs properly through the trasparent proxy but the screenshot is not done with this option enabled ( there is no ad blocking, that is shown in the page assets, neither the widget we inject in the HTML code ).

soulgalore:
HI @daehin 
thanks for reporting! Do you have time to look into this yourself? Trying to finish off the new version (4.0) the coming months and will try to focus on this. The screenshot is done today using phantomjs in a specific script, could be that it misses the proxy option.

Best
Peter

soulgalore:
Closing old 3.x issues. Please reopen if you this still is an issue for 4.0.

PR #803, opened 2016-01-17
WIP: Add ability to use proxy configuration for screenshots

beenanner:
@daehin this needs to be tested before merging and is in progress. For: https://github.com/sitespeedio/sitespeed.io/issues/791


beenanner:
Should I add support for the headless browsers as well @soulgalore? Looks like that would require some additional work in the lib/headless/scripts/screenshot.js.

PR #829, opened 2016-03-11
Removing missed merge conflict from README.md

beenanner:


tobli:
Thanks for spotting this, I resolved it in 83ff814. The HELP reference and screen shots were for the current master, so I removed those.

PR #902, opened 2016-04-27
Chrome throws error on sitespeed results due to about:blank captured as domain in HAR

beenanner:
Temp fixing issue where about:blank is being reported as a domain and causing an error in sitespeed when running with -b chrome option. Most likely an issue that can be pushed over to browsertime. Adding this patch for now to allow sitespeed to continue to work for chrome.

### DIFF
--- a/lib/plugins/domains/aggregator.js
+++ b/lib/plugins/domains/aggregator.js
@@ -36,6 +36,11 @@ function getDomainEntry(domainName) {
 module.exports = {
   addToAggregate(har) {
     har.log.entries.forEach((entry) => {
+        // Temporarily exclude new window's about:blank for Chrome until it
+        // can be excluded from browsertime HAR results
+        if(entry.request.url == "about:blank") {
+            return;
+        }
       const domainName = parseDomainName(entry.request.url),
         domain = getDomainEntry(domainName);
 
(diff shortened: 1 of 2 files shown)


soulgalore:
Hey @beenanner interesting, how can I reproduce? I haven't seen this when I'm testing. I'll merge it just want to know how I can get the same :)
<pre>
$ bin/sitespeed.js https://www.sitespeed.io/ -b chrome
</pre>
works for me.

beenanner:
I was able to consistently reproduce with http://www.cnet.com/ on my Linux machine. I'll try sitespeedio to see if it's OS related.

beenanner:
So after testing this morning on the main 4.0 branch(excluding my change) it doesn't seem that I can reproduce the issue with 
```
bin/sitespeed.js -b chrome http://www.cnet.com/
```
I'm questioning now whether it was a caused by an ad/third party script loaded on the site. Maybe we don't merge this just yet and see if the issue crops up again in the near future? The other strange thing though was I couldn't reproduce it using firefox while the issue occurred in chrome. >_<

soulgalore:
I'm dropping this PR for now, lets see if we can get this to happen again.

Issue #908, opened 2016-04-30
Improve test setup for graphite/influxdb/grafana


beenanner:
I'm going to look into setting up a docker-compose file for running these 3 services it seems on my linux machine they aren't able to talk to one another without additional work. It would be nice to just run `cd tools;docker-compose up` for development.

soulgalore:
yes sounds good :)

tobli:
Fixed now, thanks @beenanner!

PR #910, opened 2016-05-01
#908 - Adding docker-compose.yml for testing grafana, graphite, and influxdb

beenanner:
Added docker-compose.yml for #908 

Mac steps
```
cd tools
docker-compose up -d
docker-machine ls
```
Note IP address from docker-machine command and use that for the url in grafana's data source setup.

### DIFF
--- a/tools/docker-compose.yml
+++ b/tools/docker-compose.yml
@@ -1,0 +1,28 @@
+version: '2'
+services:
+    grafana:
+        image: grafana/grafana
+        depends_on:
+            - graphite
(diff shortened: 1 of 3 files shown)


beenanner:
I was getting an issue with using the service name alias if I can fix that the name of the service could be used during the datasource setup in grafana. For example to setup the graphite datasource the URL would just be http://graphite:8080

beenanner:
Referencing the other containers within grafana should work now I had to expose the ports.

Issue #915, opened 2016-05-05
3rd party vs 1st party for sitespeed 4.0


beenanner:
For sitespeed 4.0 it would be a nice feature to be able to configure what requests are considered 1st party vs everything else which would be third party so we could present statistics around those configurations. Possible have some sort of domain whitelist or JSON file for naming a group of domains would be nice.

soulgalore:
Yep I agree. One fat regular expression? Easiest way should be to specify what's 1st party right and then treat the rest as third party. 

What kind of metrics do you want to see? requests & size. Timings?

beenanner:
Yeah it would be nice to possible pre-calculate these request counts, size total, etc. before sending them over to a time-series db for storage so that they wouldn't have to be calculated on the fly as is currently done in 3.x

tobli:
Request count per domain is in by default now (total for all pages).

soulgalore:
Only missing now is the summary in https://github.com/sitespeedio/sitespeed.io/blob/4.0/lib/plugins/coach/pagexrayAggregator.js

soulgalore:
@beenanner you can try it out now and see if it works as you want. You need to supply a regex that categorize a URL as a firstParty, all other URLs will be third.

beenanner:
@soulgalore look good everything seems to be working as expected. Great addition!

beenanner:
Everything looks good to me closing this out. Thanks guys!

soulgalore:
I will just add something in the front end  HTML.

guifreitag:
Is there a way to provide first party regex when using url's in a separate file?

soulgalore:
Hi @guifreitag yep that should just work out of the box? `sitespeed.io myFile.txt --firstParty ".*sitespeed.*"` but the regex is the same for all URLs.

Best
Peter

PR #917, opened 2016-05-06
Flatten domain names for graphite

beenanner:


tobli:
Hi!
Better to keep the source data untouched. The proper place to fix this issue would be in flattenMessage, since that's what graphite and influx uses to construct metrics data. We need to call toSafeKey on the key before adding it to the key path at https://github.com/sitespeedio/sitespeed.io/blob/4.0/lib/support/flattenMessage.js#L58. That should fix all funky key values.

beenanner:
Thanks @tobli easy enough I'll update and test today!

tobli:
Sweet. I didn't test it, so I hope it works.
A unit test for flattenMessage would be amazing, but not needed for merging.

beenanner:
Updated so it works via flattenMessage.js, but needs cleanup and a unit test.

tobli:
I was just referring to changing key to toSafeKey(key) in joinNonEmpty([keyPrefix, key], '.'). Replacing dots, spaces etc in keys isn't specific to domain stats, but a general concern.

More generally there's a contention between an optimal data structure for html generation and metrics collection. For html there's often a specific template with access to js for logic. Arrays are no problem, and we want to retain the original data (e.g. domain names) to render on the page. Metrics data is much more constrained, and we need to make sure all strings are valid key path segments. I think we should keep the "source" format at the producer as agnostic as possible, even if it means a slight transform at the consumers (html, graphite etc). Given that, the current summary format for assets and domains are probably not ideal. I have a patch for domains coming up.

The final piece of the puzzle are metrics filters. Currently everything from domains is filtered out, i.e. excluded by default. See https://github.com/sitespeedio/sitespeed.io/blob/4.0/lib/plugins/domains/index.js#L13. Maybe we should store number of requests per domain as a default metrics. Totoal transfer size might be a candidate as well. There's no cli or api support yet for changing the filters, we need to figure out a good strategy for that. Right now the filter is defined at the source. Maybe it should belong to the metrics plugin instead, I'm not sure. What would be more natural, --domains.filter... or --graphite.filter.domains... ?

tobli:
I actually included the toSafeKey fix in d00261b82e2b366f4713086120b010b1ea74cff6 as a part of a bigger reshuffling of domain data + adding request count per domain to graphite. Sorry for overwriting this PR. Great catch to find that domain names in keys needed escaping, too bad it was caught up in a bigger mess with a bad data structure. Should be better now, I hope.

beenanner:
No worries @tobli! I knew I was missing something. I was staring at the second "key" on that line you posted and couldn't figure out why the domain names were not going through that. Thanks for the explanation! I agree the consumer seems to be the best place to apply filters since multiple consumers could request/use different filters (e.g. send all domain data to graphite, but not to influxdb). I was thinking the json files would be the source of configuration for all metrics for the consumers. This would make it a simple source for the user to override any default configs. Would it make it cleaner to leave all the producers returning all metrics and have a default json configuration filter at the consumer level for each with the basics that you mentioned? The user could then just pass in a new JSON config file path to override the default configs that are include out of the box. I guess this long winded answer with my thinking has my opinion falling more inline with the metrics plugin handling the filtering :-p

beenanner:
> Maybe we should store number of requests per domain as a default metrics. Totoal transfer size might be a candidate as well. 
@tobli I think storing overall metrics for each domain would be a great idea for default metrics if we can easily calculate them at the producer level. I think it would help out with #915 a lot and would allow for some dashboards I had in mind to be created without sending all the domain data to influxdb or graphite.

PR #925, opened 2016-05-14
Aliasing crawler depth to '-d' for easy of use.

beenanner:
Seemed like this would be a good candidate for aliasing as it would most likely be a commonly used option.

### DIFF
--- a/lib/support/cli.js
+++ b/lib/support/cli.js
@@ -74,7 +74,8 @@ module.exports.parseCommandLine = function parseCommandLine() {
     /*
      Crawler options
      */
-    .option('crawler.depth', {
+    .option('d', {
+      alias: 'crawler.depth',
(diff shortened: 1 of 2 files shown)


lib/support/cli.js
```diff
@@ -254,6 +255,11 @@ module.exports.parseCommandLine = function parseCommandLine() {
     }
     // End hack
 
+    // FIXME: Find a solution that would work for all aliasing
```
soulgalore: Thinking maybe we could find a solution for when we parse which plugins that will be used or @tobli do you have a nice solution lined up for it?

tobli: As a matter of fact... I have an idea of how to do it. But we can leave this in for now.

soulgalore: Lets merge it then.

PR #929, opened 2016-05-15
Unescaping <br> in page summary report for dns-prefetch

beenanner:
Notice the <br> tag was printed due to being escaped by default.


tobli:
Perfect! Would it be ok to squash the commits?

soulgalore:
I think you can use the squash button, it should be availible, however you need to choose merge first in step one (it is not 100% clear).

tobli:
Sweet! Didn't know that was available on mobile. Thanks @beenanner

beenanner:
Oops sorry meant to squash these and wondered off to bed. Good to know github has the power. 😀

Issue #932, opened 2016-05-16
Not all metrics reach Graphite


soulgalore:
There's some strange with at least <code>sitespeed_io.pageSummary.run_sitespeed_io.*.coach.advice.accessibility</code> because it never ends up in Graphite (it looks like we send it though).

soulgalore:
fixed by @beenanner

PR #935, opened 2016-05-17
Missing newline caused last metrics to be lost when sent to graphite #932

beenanner:
The missing newline after last metric was causing the last metric in every section to not make it to graphite.


soulgalore:
Thanks @beenanner !

PR #941, opened 2016-05-19
Offender converted to unordered list to match resource hints

beenanner:


soulgalore:
Great much better, thanks @beenanner

PR #946, opened 2016-05-23
Add automatic test on WebPageTest in Travis

soulgalore:
One thing missing for us today is that we don't automatically test the WebPageTest code path. Lets do that!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,13 @@ before_install:
 script:
 - npm run travis
 - bin/sitespeed.js -b firefox -n 2 http://www.browsertime.net
-- bin/sitespeed.js -b chrome -n 2 http://www.sitespeed.io
+- bin/sitespeed.js -b chrome -n 2 https://www.sitespeed.io
+- bin/sitespeed.js -b chrome -n 2 https://www.sitespeed.io ----webpagetest.key $WPTKEY --webpagetest.host www.webpagetest.org
+env:
+  global:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -27,7 +27,13 @@ before_install:
 script:
 - npm run travis
 - bin/sitespeed.js -b firefox -n 2 http://www.browsertime.net
-- bin/sitespeed.js -b chrome -n 2 http://www.sitespeed.io
+- bin/sitespeed.js -b chrome -n 2 https://www.sitespeed.io
+- bin/sitespeed.js -b chrome -n 2 https://www.sitespeed.io ----webpagetest.key $WPTKEY --webpagetest.host www.webpagetest.org
```
tobli: Specifying webpagetest.host is not needed when using the default public host.

.travis.yml
```diff
@@ -27,7 +27,14 @@ before_install:
 script:
 - npm run travis
 - bin/sitespeed.js -b firefox -n 2 http://www.browsertime.net
-- bin/sitespeed.js -b chrome -n 2 http://www.sitespeed.io
+- bin/sitespeed.js -b chrome -n 2 https://www.sitespeed.io
+- bin/sitespeed.js -b chrome -n 2 https://www.sitespeed.io --webpagetest.key $WPTKEY --webpagetest.host www.webpagetest.org
```
tobli: Specifying webpagetest.host is not needed when using the default public host.

tobli:
👍

PR #949, opened 2016-05-23
Pug is the new Jade

soulgalore:
Yep it's still in alpha but and it works nice and will probably be released before we do 4.0.


tobli:
Need to update the requires as well. I actually created a local branch with the same name. Great minds... ;-)

soulgalore:
yep sorry, npm install. maybe rename all the templates too.

soulgalore:
Lets fix this later and make sure Pug works.

PR #955, opened 2016-05-24
Lodash package updates

tobli:
Minor updates to a suite of lodash libraries.


tobli:
Purposefully didn't update the changelog for this. I'd prefer to have the changelog be just user facing changes. A constant stream of npm package upgrades will just be noise imho. For the full details, there's always git log. However I'm open to good counter arguments!

soulgalore:
git log = code is the documentation :)

I see your point that don't put all the small things in the changelog but it has helped me before to just write which packages that's been updated between releases when there's a new bug, it's much easier to backtrack if something has been introduced or not.

Issue #962, opened 2016-05-27
Unable to use 4.0 new parameter --browsertime.preTask


LiyaGuo:
Hi Peter,

Thanks for your example.:
 https://github.com/tobli/browsertime/blob/1.0/test/preposttasks/preLoginEx
ample.js 
I copied the example file as
“/Users/lguo/git/sitespeed.io/test/helpers/login.js”, and modified my
package.json to include "selenium-webdriver”, and did npm install.

However, I am still unable to use it.  I added some logs to engine.js, but
still cannot figure out why. Could you help me?

➜  sitespeed.io git:(master) ✗ bin/sitespeed.js
--browsertime.preTask=/Users/lguo/git/sitespeed.io/test/helpers/login.js
https://wikipedia.org
[2016-05-24 18:47:38] Versions OS: darwin 14.5.0 sitespeed.io:
4.0.0-alpha.1 browsertime: 1.0.0-alpha.18 coach: 0.22.3
======== temp log ======= options:
{"statistics":true,"preTask":"/Users/lguo/git/sitespeed.io/test/helpers/log
in.js","browser":"firefox","iterations":3,"viewPort":"1366x708","coach":tru
e,"resultBaseDir":"/Users/lguo/git/sitespeed.io/sitespeed-result/wikipedia.
org/2016-05-24-18-47-38"}
======== temp log ======= this.options:
{"scripts":[],"iterations":3,"delay":0,"experimental":{},"statistics":true,
"preTask":"/Users/lguo/git/sitespeed.io/test/helpers/login.js","browser":"f
irefox","viewPort":"1366x708","coach":true,"resultBaseDir":"/Users/lguo/git
/sitespeed.io/sitespeed-result/wikipedia.org/2016-05-24-18-47-38"}
[2016-05-24 18:47:38] Starting firefox for analysing https://wikipedia.org
3 time(s)
======== temp log ======= options.preTask:
/Users/lguo/git/sitespeed.io/test/helpers/login.js
======== temp log ======= runIteration: 0
[2016-05-24 18:47:41] TypeError: preTask.run is not a function
    at
/Users/lguo/git/sitespeed.io/node_modules/browsertime/lib/core/engine.js:13
0:63
From previous event:
    at
/Users/lguo/git/sitespeed.io/node_modules/browsertime/lib/core/engine.js:13
0:28
From previous event:
    at runIteration
[...]

PR #966, opened 2016-05-31
Read pre/post tasks and pass the code to Browsertime #962

soulgalore:


tobli:
Great to have this fixed! I'd suggest moving the file loading out of cli.js. Later we should have this functionality in BT and exposed as api, but this is a good start.

soulgalore:
Fixed this in BT https://github.com/tobli/browsertime/pull/145

PR #976, opened 2016-06-05
Minor tweak to only display 3rd vs 1st module in HTML output if values exist

beenanner:
Only show 1st vs 3rd summary in HTML output if it has data and show default storage location for HTML ouput in help menu


soulgalore:
Perfect thank you @beenanner

Issue #874, opened 2016-04-13
Help method for displaying URLs


soulgalore:
Today we display the full URL for an asset, check this out:
<img width="978" alt="screen shot 2016-04-14 at 00 59 35" src="https://cloud.githubusercontent.com/assets/540757/14512071/4910cf98-01dc-11e6-89df-5330c99c1b25.png">

Ooops that is quite long. We could try to do something magical like displaying the domain and the end of the filename and skip parameters? I think webpagetest has an ok way of doing it in the waterfalls, perfcascade has something like that too.

dmillerme7:
There might be a way to do this in PHP if you're using that, looks like WPT is. Unfortunately I am not a PHP person. Perfcascade is using SVG so that it a different story altogether. Using JS and jQuery we can use replace() and split() to get something similar happening. If there was a class on all those anchor tags it would make it easier. Are you going to use jQuery in v4.0?

dmillerme7:
A very scruffy jsfiddle https://jsfiddle.net/dmillerme7/v4ga3uv3/4/

soulgalore:
Hey @dmillerme7 thanks for the jsfiddle :) We will avoid JQuery but we can hack around that. We can do it in the layer when parsing the output and we can add some tests for it.  Do you have to give it a go or I can set something up.

dmillerme7:
Hi @soulgalore I'll clone the project and have a go at it. If I get something working I'll make a pull request.

beenanner:
@soulgalore / @dmillerme7 I threw together a quick hack for this issue that we should clean up as well as it definitely can be improve upon. Check it out and let me know if it gives us what we need for now.

PR #977, opened 2016-06-05
Shrink URL on Asset HTML report for any relatively long URLs

beenanner:
Quick hack to add a shortenURL for HTML output for URLs longer than 30 characters for #874. Code should probably be moved to a function and tests needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 -------------------------
 ### Fixed
 * The CSS size advice on summary page used the wrong metrics to check the color, meaning 0 bytes made it red :/
+* Shorten long URLs displayed on HTML Asset report #977
 
 ## 4.0.0-alpha5 - 2016-06-30
 -------------------------
@@ -38,7 +39,6 @@
 ### Fixed
 * Enable the coach by default.
 * Fixed broken default plugin loading.
-
 
 ## 4.0.0-alpha2 - 2016-06-07
 -------------------------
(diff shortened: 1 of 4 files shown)


soulgalore:
Cool I will test this tonight, think we could add this to more places.

soulgalore:
And you could backport it to PerfCascade, I did a super simple version there nad it can be improved + it would be cool to use the same logic.

soulgalore:
Did a quick check, I think we should aim for keeping the last part of the URL, that's the part that helps us identify the exact asset? I like how WPT does it:
screen shot 2016-06-05 at 20 22 09

What do you think @beenanner ?

beenanner:
I agree and like it as well. Looks like they first grab from the file name then trim from the domain name if they have to. Let me take a look at their implementation to verify and see what I can do.

beenanner:
Code for reference: https://github.com/WPO-Foundation/webpagetest/blob/9d1084d97765d0c15a1f5a822c60ec885522f0c6/www/waterfall.inc#L1100

dmillerme7:
Nice job!

beenanner:
If this works for everyone I can merge this into master for the next release.

Before
before


After
after

soulgalore:
yes lets do it, Ive missed this one.

tobli:
Looking good! One idea to build in this would be to remove the protocol from the url. Then indicate with https or not with an icon (e.g. a padlock). Could make the secure/not more easy to skim, regardless of the domain.

beenanner:
Good idea let me file a ticket to investigate and implement that @tobli!

PR #981, opened 2016-06-07
Empty size fixes for tablet view of pages HTML report

beenanner:
Returning N/A if nothing to display to fix issue with text overlap on tablet view of 'Requests and sizes per content type' section of pages and switching to captial unit text to clarify that it is in Bytes.

https://xkcd.com/394/


beenanner:
Before
before


After
after

PR #988, opened 2016-06-09
Adding cli option to toggle waterfall on pages summary on

beenanner:


soulgalore:
Looks great, what do you think about creating a new cli group **html** where we can put everything that is for the html output. Maybe also make it html.showWaterfallSummary?

lib/plugins/html/htmlBuilder.js
```diff
@@ -19,6 +19,7 @@ class HTMLBuilder {
     this.summaryPages = {};
     this.urlPages = {};
     this.urlRunPages = {};
+    this.showWaterfallSummary = false;
```
tobli: No need to make this an instance variable (see below).

lib/plugins/html/htmlBuilder.js
```diff
@@ -111,6 +115,7 @@ class HTMLBuilder {
       menu: 'pages',
       pageTitle: name + ' ' + url,
       pageDescription: '',
+      showWaterfallSummary: this.showWaterfallSummary,
```
tobli: Remove and pass via locals argument

lib/plugins/html/htmlBuilder.js
```diff
@@ -60,6 +61,9 @@ class HTMLBuilder {
 
   renderHTML(options) {
     log.info('Render HTML for %s page(s) ' , Object.keys(this.urlPages).length);
+    if ( options.waterfallSummary) {
+      this.showWaterfallSummary = true;
+    }
```
tobli: Just add the option to the "this._renderUrlPage" call below.

lib/plugins/html/htmlBuilder.js
beenanner: Well that is soo much easier thanks!

beenanner:
@soulgalore / @tobli  I split out the waterfall into it's own jade file and added it to the individual runs. Still need to figure out a way to target specific har logs for the individual runs. Possible just need to create a new method to extract that data from the available logs within the HTML builder to include in the template build out. Let me know what you guys think.

soulgalore:
I think we been working on the same thing or almost. If you update (I've already added the HAR on the individual page, using a function in the coach to a pick a specific page from the HAR) the code. You can move over the change I did (to just pick the HAR) and add the small fix to add the firstPaint.

PR #990, opened 2016-06-10
Start work on moving to SASS for CSS

matthojo:
## Description
This commit does most of the work to bring the CSS over to SASS. There are only small bits that still need sorting out into 'components', they are in `misc.scss` - I just wasn't sure of the best place to put them. There is still some work needed to move over all `px` values to `rem`, I've already moved over a few.

The SCSS structure will still need optimising a little and possibly a naming scheme introduced (BEM, Basscss etc..). This is something that we can develop with further discussion.

## Prominent Changes
 - Use `npm run build:css` to build the CSS files after SASS changes.
 - The `layout.jade` now only requests 1 CSS file, `index.min.css`.

## Known Issues
For some reason the page headings are not quite as bold as before, it seems to be down to differences in "HelveticaNeue" and "Helvetica Neue".


soulgalore:
Great @matthojo so do you want me to merge or do you want to fix the rest first?

matthojo:
If we merge now it will allow new branches to use SASS layout without having to be merge in - thinking HAR viewer. It's only minor changes that new doing like `px` to `rems` for example, so if you're happy with the current position it is in I'm happy for it to be merged :)
Alternatively if you would rather me finish off (maybe later tonight) the smaller details I can do.

tobli:
We should npm ignore lib/plugins/html/src so it's not included in the published package.

matthojo:
I was going to do that, but I thought the page build would then need to generate the CSS files. So to keep as close to current setup I left them in, it is something I can definitely add if you would like.

tobli:
How do we best ensure that the checked in precompiled css is kept up to date? It's can be easy to forget checking stuff in. Maybe a test that compares runs "npm run build:css" and checks that lib/plugins/html/assets/css/index.css is unchanged. Then at least Travis can help us catch that.

package.json
```diff
@@ -42,7 +42,8 @@
     "pug-lint": "pug-lint lib/plugins/html/templates",
     "test": "mocha",
     "check-licenses": "tools/check-licenses.js",
-    "travis": "npm run lint && npm run test"
+    "travis": "npm run lint && npm run test",
+    "build:css": "node-sass lib/plugins/html/src/sass/main.scss > lib/plugins/html/assets/css/index.css && cleancss -o lib/plugins/html/assets/css/index.min.css lib/plugins/html/assets/css/index.css"
```
tobli: We need to add sass as a dev dependency for this, right?

matthojo: :O Correct, I thought I added this in already! Good spot, I will push a PR now.

tobli: No worries. You could add " && npm run build:css" to the travis script. It's always helpful to have Travis execute stuff to catch missing dependencies.

soulgalore: added sass :)

Issue #952, opened 2016-05-24
Add screenshots


soulgalore:
We have screenshots in Browsertime, so we should just configure to use it + when we use Firefox we should crop the image to fit the current viewport.

soulgalore:
We need to find a plain nodejs library to crop the image from Firefox + find a place to show the image on summary page + each individual run.

PR #1014, opened 2016-06-20
Collect screenshots from Browsertime

soulgalore:
Chrome is easy, Firefox needs to crop the image to the current viewport.
There's also changes to the layout, to make the image fit in better (and
we can do this much better).

See this as a first move for #952 and let us improve.


soulgalore:
We probably should make this configurable: Taking a screenshot on a site like expressen.se takes quite long time with Firefox.

PR #1046, opened 2016-07-08
Sync'ing up menu text on page summary and run summary

beenanner:
@soulgalore may be a good idea before 1.0 we split some of these(menu, screenshot, page summary, etc) out into partials that we can include to avoid duplication of HTML and keep things consistent. Not too important though but would be an easy ticket for someone that wants to contribute to pick up.


soulgalore:
@beenanner sounds good!

Issue #1003, opened 2016-06-16
"Pages" link to specific page Coach data doesn't have styles on Windows


rschoessler:
```
Given I run a sitespeed test on a web page
And I am viewing the HTML results
When I go to the Pages link in the header menu
And I select a page URL to view the page details
Then I should see a formatted page with CSS styles
```
When I view on Mac, styling works fine.
When I view on Windows, the page with the "Coach" data shows no styles.

>file:///C:/sitespeed/sitespeed.io/sitespeed-result/myurl.com/2016-06-16-13-30-11/pages/myurl.com/index.html

![coach](https://cloud.githubusercontent.com/assets/3880296/16126780/9c2c2f84-33c7-11e6-90b3-1cb03e7f2cdc.png)

soulgalore:
hey @rschoessler thanks for reporting, could you do a view source and just check what the link to the CSS looks like?

Best
Peter

rschoessler:
So, if I edit index.html in 


Sure, here's what I see on Windows:

`<link rel="stylesheet" href="../css/index.min.css">`

On Mac I see:

`<link rel="stylesheet" href="../../css/index.min.css">`

if I update the .html manually on Windows to follow the same path as Mac, I get the styles.

Also, the Images on that page show the same issue and are fixed when updating the path.

soulgalore:
I guess this could be a // on windows or something like that. Lets create storage manager tests, so we can verify that it really works.

rschoessler:
Hi guys...it looks like the links in the header navigation on the .../pages/site_url/index.html are also broken.  They are looking for the pages in the /pages directory instead of the root folder for that particular run.

soulgalore:
Yes, I think @tobli has a fix for this and soon he can commit it so you can test that it works.

tobli:
Fixed via #1084

PR #1084, opened 2016-07-26
Correct way of creating relative urls in html site

tobli:
- rename relativePathFromUrl to pathFromRootToPageDir to make it more obvious what it does.
- join relative url path with / instead of using the path module (should make paths work on all OSes).
- remove path parameter for url run pages, since it wasn't used.
- make StorageManager constructor signature follow node conventions (options last) and add some StorageManager tests.

### DIFF
--- a/lib/plugins/html/htmlBuilder.js
+++ b/lib/plugins/html/htmlBuilder.js
@@ -27,7 +27,7 @@ class HTMLBuilder {
 
   addUrl(url) {
     this.urlPages[url] = {
-      path: this.storageManager.relativePathFromUrl(url),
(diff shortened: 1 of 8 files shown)


tobli:
Since I've fixed some path separator related issues, I'm hoping this could fix #1003. Would love some help verifying on Windows.
Ping @beenanner, @rschoessler

beenanner:
@tobli will try to test it tonight!

beenanner:
@tobli verified this branch fixes the issue with #1003!

tobli:
👍

Issue #1148, opened 2016-08-22
Better defaults for window size


soulgalore:
Lets set  1280x1024 and make sure ffmpeg also works with that in the Docker container.

See https://github.com/sitespeedio/sitespeed.io/issues/1143.

soulgalore:
There's no good way to fix this until #1143 and upstream in Firefox. The user experience will be bad with tests breaking, so ... should we just set it to smallish when you run Firefox until its fixed upstream?

soulgalore:
It's fixed in the new Docker container, lets wait with the change til the fix for max window size  lands in Marionette stable.

Issue #1143, opened 2016-08-18
Requested size exceeds screen size error [Firefox]


soulgalore:
We have this upstream thing mozilla/geckodriver#56

What happens right now is that when you open FF with larger window than the screen, we get an error and fails. It's fixed but not released in FF so we should probably change the default width we set (or at least make sure the Docker container works out of the box).

beenanner:
https://www.x.org/archive/X11R7.6/doc/man/man1/Xvfb.1.xhtml The default is 1280x1024x8 which is what docker would use correct?

soulgalore:
It's 1366x768x16 but lets change that and change default in sitespeed.io to the same!

soulgalore:
And we need to trim ffmpeg so it's some work to make sure it works fine.

soulgalore:
In the new Docker file we have better default for xvfb, then later on when the fix lands in Firefox/Marionette/Gecko, lets change at our end.

soumyamulu:
Do we know whether the fix for the window size error is done or rolled back. I am currently using FF 48+ and Selenium(2.53.1, 3.0.0-beta(123)) all give me the same exception unless i set the window size to 1024X768. This is breaking my tests because the elements are out of the window scope.

According to this bug 1280101 they rolled back the changes? Anyone else seeing this problem?

soulgalore:
hey @soumyamulu I think they fixed it in Marionette but it's waiting to get out in coming Firefox.

soumyamulu:
Thank you for the clarification. Till then do we have any workaround for this problem? 
for some reason I can no longer specifiy the binary path as the Firefox developer edition in selenium 3.0+ it always seems to use the real FF. I was wondering if  I can see this fix in the dev version?

soulgalore:
Do you need a workaround for sitespeed.io or selenium? :)

I think (if I remember correctly) that we also need a new version of the Geckodriver, best to check with the Firefox team.

tobli:
@soumyamulu The binary path config was changed quite recently in selenium git, so can't give any general advice. You need to check the source code/git logs for whatever exact revision your running. And also make sure you run a corresponding version of geckodriver. Should work, but haven't checked myself. Note that the latest change are not yet in Browsertime/Sitespeed.io, since they're not released to npm yet.

soumyamulu:
Thank you for the clarification @soulgalore  and @tobli 
[...]

PR #1151, opened 2016-08-23
Increasing default screen size #1143, #1148

beenanner:
Will update the changelog once we've confirmed this is working as expected.


soulgalore:
I think it don't work on Travis either since we cannot set the browser at 0.0 position (will come later in gecko or if it was selenium) but you can just increase the screen size in Travis I think it will pass: 
https://github.com/sitespeedio/sitespeed.io/blob/master/.travis.yml#L26

Issue #1145, opened 2016-08-19
Different summaries for different files (and rename summary field)


soulgalore:
We should aim to fix this before we release the first beta. 

If I run like this:
<code>bin/sitespeed.js test.txt test2.txt -n 1</code>
And the test.txt file contains https://www.sitespeed.io and the second one contains https://github.com we will get the following categories of keys:
<pre>
sitespeed_io.pageSummary.www_sitespeed_io._.
sitespeed_io.pageSummary.github_com._.
sitespeed_io.summary.test_txt.
</pre>

To make it easy to test multiple sites, it should result in something like:
<pre>
sitespeed_io.pageSummary.www_sitespeed_io._.
sitespeed_io.pageSummary.github_com._.
sitespeed_io.summary.test_txt.
sitespeed_io.summary.test2_txt.
</pre>

And best case scenario, we could add some logic that if the file only contains one domain, we could use that instead of the name of the file.

soulgalore:
Discussed this with @tobli and we could collect starts per domain + a total stats (then choose if we should send the total or not). That will probably solve all use cases except how to go from a summary of a domain to the pages for that domain, but maybe we can solve that later.

soulgalore:
I'll start on this and push a branch when ready, need help to test when it is ready.

beenanner:
Let me know and I can help :-)

soulgalore:
You sure can :) I'm not sure about the implementation, wanna do it as simple as possible bit haven't got a clean idea yet.

soulgalore:
I've pushed a branch now. A couple of things: We call it group(s) now but always group by domain as now (thanks @tobli for the idea). We collect summaries per group and per total of all groups. We don't send the total to graphite, but we use it in the summary page for the HTML output.

Room for improvement in the future: 
* Group pages per group
* Add new HTML if you have multiple groups (like the old --sites)
* more?

soulgalore:
But I think (if this works and you @beenanner and @tobli feels it's ok) we can have it for the first beta.

PR #1153, opened 2016-08-23
Lets group summaries (by domain for now) #1145

soulgalore:


soulgalore:
Lets fix WebPageTest & GPSI also ...

soulgalore:
We don't aggregate GPSI yet so it is ok to leave it out.

soulgalore:
It's ready for testing and input :)

PR #1212, opened 2016-09-20
Configure extra plugin dir

soulgalore:


soulgalore:
lets use @tobli simpler solution

Issue #1218, opened 2016-09-24
pug template cache is never a 'hit'


moos:
Due to the async nature of the render* calls, the template cache is never hit. 

```shell
$ bin/sitespeed.js -m3 -n1 http://0.0.0.0:8082
[2016-09-24 00:27:18] Versions OS: darwin 14.5.0 nodejs: v6.5.0 sitespeed.io: 4.0.0-beta.1 browsertime: 1.0.0-beta.3 coach: 0.25.0
[2016-09-24 00:27:19] Starting chrome for analysing http://0.0.0.0:8082 1 time(s)
[2016-09-24 00:27:19] Testing url http://0.0.0.0:8082 run 1
[2016-09-24 00:27:26] Starting chrome for analysing http://0.0.0.0:8082/ 1 time(s)
[2016-09-24 00:27:26] Testing url http://0.0.0.0:8082/ run 1
[2016-09-24 00:27:32] Starting chrome for analysing http://0.0.0.0:8082/.git/ 1 time(s)
[2016-09-24 00:27:32] Testing url http://0.0.0.0:8082/.git/ run 1
[2016-09-24 00:27:39] Render HTML for 3 page(s) 
assets.pug cached? false
toplist.pug cached? false
index.pug cached? false
detailed.pug cached? false
domains.pug cached? false
pages.pug cached? false
help.pug cached? false
url/index.pug cached? false
url/index.pug cached? false
url/index.pug cached? false
  compiled assets.pug 1
    render assets 1
  compiled index.pug 2
    render index 2
  compiled toplist.pug 3
    render toplist 3
  compiled detailed.pug 4
    render detailed 4
  compiled pages.pug 5
    render pages 5
  compiled domains.pug 6
    render domains 6
  compiled help.pug 7
    render help 7
  compiled url/index.pug 8
    render url/index 8
  compiled url/index.pug 9
    render url/index 9
  compiled url/index.pug 10
    render url/index 10
url/run.pug cached? false
url/run.pug cached? false
url/run.pug cached? false
  compiled url/run.pug 11
    render url/run 11
  compiled url/run.pug 12
    render url/run 12
  compiled url/run.pug 13
    render url/run 13
[2016-09-24 00:27:43] HTML stored in /Users/dev/git/sitespeed.io/sitespeed-result/0.0.0.0/2016-09-24-00-27-18
[2016-09-24 00:27:43] Finished analysing http://0.0.0.0:8082
```
[...]

PR #1220, opened 2016-09-26
Process pug synchronously for better memory management - fixes #1219

### LINKED ISSUE #1219: Sitespeed becomes a memory hog for large # of urls
Possibly cause of #1179.  Run with 5 urls (1 run each) consumes ~370MB.  With 25 urls, about 730MB.  Here's the memory allocation map for the 25-url run.  As can be seen about 700MB (between 50MB and 750MB ticks on the x-axis) is due to the HTML render section.

screen shot 2016-09-23 at 11 58 46 pm

Memory is rarely de-allocated -- largely due to the async nature of the code.

moos:

Fixes #1219 and #1218.

There is no clear benefit in rendering async as each process is roughly of equal time -- so I don't think we losing anything going sync -- and gaining much better memory management.

### DIFF
--- a/lib/plugins/html/renderer.js
+++ b/lib/plugins/html/renderer.js
@@ -1,11 +1,7 @@
 'use strict';
 
 const pug = require('pug'),
-  fs = require('fs'),
-  Promise = require('bluebird'),
   path = require('path');
-
-Promise.promisifyAll(fs);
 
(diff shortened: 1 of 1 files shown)


soulgalore:
thanks @moos for taking the time to investigate and fix this 🎯

PR #1227, opened 2016-10-01
Budget

soulgalore:
This starts to look like something we can merge as a first version. One thing: I'm not sure about having ok/failures separated, we could probably just keep em per URL and add a status field.

Junit looks like this:
screen shot 2016-10-01 at 23 06 53
screen shot 2016-10-01 at 23 06 43
screen shot 2016-10-01 at 23 06 35

And TAP:
screen shot 2016-10-01 at 23 07 32

Clicking on a failed test in TAP gets a 404 though?


moos:
+1 much needed for Jenkins integration.

Issue #1223, opened 2016-09-27
Give the slack plugin some love


soulgalore:
When we added the dataCollector plugin, it's really easy to slack some interesting metrics. I'm thinking we can make it configurable so we can send a message per URL and/or per summary (and include the metrics we think it's most important + look at Browsertime new info log where we output metrics (@moos), that can kind of make into slack.

soulgalore:
it got some love.

PR #1228, opened 2016-10-04
Send summary and or run data to Slack #1223

soulgalore:
First version of sending valuable metrics to Slack.


soulgalore:
One thing I need to fix, I've hard coded firstPaint and that will not be good for FF, lets fix that first before merge.

lib/plugins/slack/index.js
```diff
@@ -3,41 +3,52 @@
...
-    if (!options.channel.startsWith('#'))
-      options.channel = '#' + options.channel;
+    const slack = new Slack(options.slack.hookUrl);
+    let logo = "https://www.sitespeed.io/img/slack/sitespeed-logo-slack.png";
 
-    let text = 'Sitespeed done!';
-    if (errors.length > 0)
-      text += util.format(' (%d) errors', errors.length);
+    if (!options.slack.channel.startsWith('#')) {
+      options.channel = '#' + options.channel;
```
tobli: This should be options.slack.channel = …

Issue #1062, opened 2016-07-19
Posting data to Elasticsearch rather than graphite


softwareklinic:
Team,

Are we planning to push the data to elasticsearch in addition to graphite.... since that can be used to further tie the data to other data sources such as dynatrace, web logs, tealeaf etc... and then do a whole 9 yards of analysis.

Just a thought... if we are planning to support it..

beenanner:
Hey @softwareklinic,

Sounds like a great idea! For the initial release though we are looking to support only graphite and influxdb initially.

ftore:
Hi guys, I am interested on implementing this feature. Is anybody working on it already?

beenanner:
@store I don't believe so. That's great we have a slack channel you can join and we can help and discuss implementation!

softwareklinic:
I would join as well


On Fri, May 26, 2017 at 9:24 PM Jonathan Lee <notifications@github.com>
wrote:

> @store I don't believe so. That's great we have a slack channel you can
> join and we can help and discuss implementation!
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_sitespeedio_sitespeed.io_issues_1062-23issuecomment-2D304417336&d=DwMCaQ&c=udBTRvFvXC5Dhqg7UHpJlPps3mZ3LRxpb6__0PomBTQ&r=jE0nj8mYWFLxOa6KDmmXBmmwqzThJ1FTJw0CAx3FXrI&m=7jKqVBOIB7DKjhkjZiYHqPQ1xUl0H4s7NKaeDod3ojI&s=hm5azvlrtLnfXd7S0-BW-PCACCr3pkMeEeyLsle5Nhg&e=>,
> or mute the thread
[...]

PR #1239, opened 2016-10-10
WIP: Inital take on adding a plugin for Elasticsearch

orjan:

- [x] I'm making a big change or adding functionality so I've already opened an issue proposing the change to other contributors, so I got feedback on the idea before took the time to write precious code: #1062 
  This pull request is not intended to merge in its current state, but more to gather requirements how to write a proper Eleasticsearch plugin.
### Description

This pull request is just about to get a feeling about what it would require to write a proper plugin for sending data to Elasticsearch. The Eleasticsearch plugin is ported straight from the Graphite plugin and some questions has emerged that I think can be valuable before trying implement a proper plugin that might be used as requirements for #1062.
1. Removal of data
   In the Graphite plugin we're removing lots of data, with:

```
    if (message.group === 'total')  {
      return;
    }

    const filteredMessage = filterRegistry.filterMessage(message);
    if (isEmpty(filteredMessage.data)) {
      log.info(`Removed ${filteredMessage.type} since data is empty after filtering.`);
      return;
    }
```

Will this hold true for the Elasticsearch plugin as well?
1. Destructing of messages
   At the moment I've just saved the messages as they are in separate indices but this can be problematic for aggregation of keys e.g. listing all advices together over time. Destructing of message will be even more important if we're not removing data in 1. There are also some fields in the messages that's a bit redundant e.g. `_type` and `type` and all data is prefixed with `data`, that might be a good thing but needs clarification.
2. Indices and Type
   Elasticsearch is a little bit different to Graphite/Influx with indices and types and different data types cannot overlap within the same index. In this PR I've just used a specific index per elastic and sitespeed data type.
3. Default pattern for index names
   I've just applied index name pattern similar to the one for logstash to enable search over time, retention and searching over all data types, but this needs to be able to be configured and possible evaluated depending on 2 and 3 above.  

```
`sitespeed-${this.timestamp.format("YYYY.MM.DD")}-${message.type.toLowerCase()}`
```
1. Mappings
   If we should add support we should probably add a dynamic template as a sensible default, to make sure that the indices are mapped correctly.

If you are interested and if we can find answers to the questions above I'm willing to continue the work and add tests for the implementation!
### How to run

First of all we'll need to start Elasticsearch and Kibana

```
cd tools/elasticsearch
docker-compose up
```

Then we'll need to activate the elasticsearch plugin and execute sitespeed.

```
bin/sitespeed.js --plugins.load elasticsearch --elk -n 2 -d 1 https://www.sitespeed.io/
```

Spiked implementation of #1062

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules
 .vscode
 sitespeed-result
 configs
+.idea
--- a/lib/plugins/elasticsearch/datagenerator.js
+++ b/lib/plugins/elasticsearch/datagenerator.js
@@ -1,0 +1,18 @@
+const log = require('intel');
+
+module.exports = dataGenerator;
+
+function dataGenerator(type) {
+  const filters = {
+  };
+
+  if (type in filters) {
+    return filters[type];
+  }
+
+  return (sender, timestamp, message) => {
+    log.debug(`No custom data generator is available for ${type}`);
+    sender.send(message);
+  }
+}
+
--- a/lib/plugins/elasticsearch/index.js
+++ b/lib/plugins/elasticsearch/index.js
@@ -1,0 +1,39 @@
+const path = require('path');
+const Sender = require('./sender');
+const filterRegistry = require('../../support/filterRegistry');
+const isEmpty = require('lodash.isempty');
+const log = require('intel');
+const throwIfMissing = require('../../support/util').throwIfMissing;
+const dataGenerator = require('./datagenerator');
+
+module.exports = {
+  name() {
+    return path.basename(__dirname);
+  },
+  open(context, options) {
+    throwIfMissing(options.elasticsearch, ['host', 'port'], 'elasticsearch');
+
+    const opts = options.elasticsearch || {};
+    this.sender = new Sender(opts.host, opts.port, context.timestamp);
+  },
+  processMessage(message) {
+    if (!(message.type.endsWith('.summary') || message.type.endsWith('.pageSummary')))
+      return;
+
+    // we only sends individual groups to Elastic
+    // NOTE: this is ported straight from the Graphite plugin
+    if (message.group === 'total')  {
+      return;
+    }
+
+    const filteredMessage = filterRegistry.filterMessage(message);
+
+    if (isEmpty(filteredMessage.data)) {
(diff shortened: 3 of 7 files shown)


soulgalore:
Hey @orjan looks super cool!  Me, @tobli or @beenanner  will get back the coming days with some feedback, trying to finish off a couple of things before the last(?) beta. 

Best
Peter

soulgalore:
Hi @orjan ,
I'll start and then can @tobli fill in if I'm missing something.

First:  An Elastic search plugin is great and it's super that you took a first stab at this, specially since we haven't added any docs yet about the structure. One thing that we have been thinking is that we will try to keep plugins as individual npm modules, to try to minimize the overhead of having too much included inside of sitespeed.io . That will make it easier for us as a small team to maintain it, so if it's ok for you, it can live in your repo and then we can link to it from the documentation? It will be one extra step for a user to use the plugin but it will make life much easier for us. We are actually thinking about moving out some (or one) of the existing plugins before launch, let see.

>1. Removal of data In the Graphite plugin we're removing lots of data, with:

One thing that we want to do "right" with 4.0 is not to overload Graphite with metrics. We did that in 3.x where we send too much data, so now we try to focus and limit the amount. The group *total* is data summarized for all different groups (=domains) so practically if you test one site, this will be the exact same metrics. If we test multiple domains the *total* is the summary of all the tested domains and in Graphite we can calculate those ourselves. We have that layer to be able in the future test multiple sites in one go (the HTML for 4.0 doesn't handle that, we had that functionality in 3.x).  Was this understandable?

> Destructing of messages
> type & _type

Need to look into that. For the prefixed *data* I think that's something that we have had since the first rewrite and never really thought of but we could just remove that, @tobli?


 > Indices and Type

Haven't worked with Elastics search, do what you feel is the best :)

beenanner:
> Mappings If we should add support we should probably add a dynamic template as a sensible default, to make sure that the indices are mapped correctly.

Agreed I believe if we want to using multiple sitespeed-<date>-<type> indices we would need default templates to add back some of the mappings that come with elasticsearch's default template as well as making sure fields are mapped to the correct types. I have a little experience in ES so I can help when I have time if you need it.

soulgalore:
I think we wanna keep this out of sitespeed and have it as a plugin. @orjan let me know if you need any help with that.

Issue #1240, opened 2016-10-11
Move site to docs within the project


beenanner:

soulgalore:
Ready to close it @beenanner ?

beenanner:
Yep we can close this!

PR #1241, opened 2016-10-11
Moving documentation to github pages #1240

beenanner:
You can see it what it will look like here: https://beenanner.github.io/sitespeed.io/. Once merged it should become available here: https://sitespeedio.github.io/sitespeed.io/ until we are ready to go live and then we can update the CNAME to point https://www.sitespeed.io/ to it.


soulgalore:
I've done some work in the [4.0 branch](https://github.com/sitespeedio/doc.sitespeed.io/tree/4.0) is this from that branch? I'm thinking maybe we can just continue there until it's in a releasable state? 

Do we have a fix for setting cache headers? I'm just thinking about what we will solve here? If we push, we will still have the current Fastly setup and need to configure caching in the front layer? And then if do a push, we still need to manually empty the cache at Fastly?

beenanner:
Yes I'll base any work off that branch if we move forward. I'm cool with working out of that docs.sitespeed.io repo for now. I'll commit to that one going forward. Does the current site require a similar manual purge at fastly? I need to look more closely at the github pages' headers. I did find this info on usage. It may purge on commit, but as you said would require a fastly purge if we are going through them. Traffic limitations is a different discussion, but if we hit that we have a good problem. :-)

https://help.github.com/articles/what-is-github-pages/#usage-limits

 I think the pros would be no server to manage and the docs would live with the code, so a pull-request for changes could also include updated documentation vs an additional pull-request via another repo. Cons: Using github pages will severely limit what we can control (i.e. cache-control headers)

soulgalore:
yes you are right hanging the docs in the same commit is good. so I maybe over sensitive to have a "perfect" setup for the docs.

beenanner:
Updated the docs with the remaining pages from the 4.0 branch @soulgalore. You should be able to check it out here: https://beenanner.github.io/sitespeed.io/. Let me know if I missed anything or we want anything added in the initial merge.

soulgalore:
I think I could have missed some changes in the CSS, the pre/code doesn't look right? But we can fix that later. There's docs missing in /3.x/ if you down a path. I think we should keep that for a while for people who don't do the switch immediately.

soulgalore:
We are missing the precious /img/404c.png and maybe need a hack for getting the 404 page to work.

beenanner:
@soulgalore moved the 404 page to a markdown file so everything should work there now.

https://beenanner.github.io/sitespeed.io/404test

Issue #1234, opened 2016-10-06
Is Slack integration fully functional?


rschoessler:
Hi...I receive the following error when doing a simple slack example.  Am I doing something wrong?

```
c:\sitespeed\sitespeed.io\bin>node sitespeed.js http://www.google.com --slack.hookUrl http
s://hooks.slack.com/services/.../.../...
[2016-10-06 07:41:06] Versions OS: win32 6.1.7601 nodejs: v4.4.4 sitespeed.io: 4.0.0-beta.
4 browsertime: 1.0.0-beta.6 coach: 0.27.1
[2016-10-06 07:41:08] Starting chrome for analysing http://www.google.com 3 time(s)
[2016-10-06 07:41:08] Testing url http://www.google.com run 1
[2016-10-06 07:41:15] Testing url http://www.google.com run 2
[2016-10-06 07:41:21] Testing url http://www.google.com run 3
[2016-10-06 07:41:27] Render HTML for 1 page(s)
[2016-10-06 07:41:28] HTML stored in c:\sitespeed\sitespeed.io\bin\sitespeed-result\www.go
ogle.com\2016-10-06-07-41-06
Warning: a promise was rejected with a non-error: [object Object]
    at runOptionalFunction (c:\sitespeed\sitespeed.io\lib\sitespeed.js:38:6)
    at c:\sitespeed\sitespeed.io\lib\sitespeed.js:104:32
From previous event:
    at c:\sitespeed\sitespeed.io\lib\sitespeed.js:104:16
    at processImmediate [as _immediateCallback] (timers.js:383:17)
From previous event:
    at c:\sitespeed\sitespeed.io\lib\sitespeed.js:89:12
From previous event:
    at Object.run (c:\sitespeed\sitespeed.io\lib\sitespeed.js:86:8)
    at c:\sitespeed\sitespeed.io\bin\sitespeed.js:25:22
From previous event:
    at Object.<anonymous> (c:\sitespeed\sitespeed.io\bin\sitespeed.js:24:4)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Function.Module.runMain (module.js:441:10)
    at startup (node.js:139:18)
    at node.js:968:3

Warning: a promise was rejected with a non-error: [object Object]
    at c:\sitespeed\sitespeed.io\lib\sitespeed.js:104:16
[...]

PR #1242, opened 2016-10-11
Should only check the slack channel prefix if the channel is defined.

orjan:
### Description
Since the default channel has been removed we cannot directly check if the channel starts with `#`, we'll need to make sure that the channel is defined before.

#1234


soulgalore:
thanks @orjan !!!

PR #1244, opened 2016-10-13
Remove additional slash to rootPath in pugs

orjan:
### Description
The rootPath ends with a trailing slash and adding an extra slash in the pugs will create an invalid url for some web servers e.g. S3

There's a unit test that asserts that the rootUrl should end with a slash and I'm assuming that it's correct. I've removed the appending slash from the locations that I can find.

screen shot 2016-10-13 at 19 23 00


soulgalore:
Good catch, thanks @orjan !

PR #1259, opened 2016-10-22
Don't run as root inside Docker

soulgalore:
Been following http://jdlm.info/articles/2016/03/06/lessons-building-node-app-docker.html  to not run as root inside the container.  I've seen when I run on Ubuntu not as root, that the owner of the files in sitespeed-result is ... root!  I've tested this locally on my Mac but not on Ubuntu, could you try it out @beenanner ?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,21 @@
 FROM sitespeedio/webbrowsers:firefox-49.0-chrome-54.0
 
-RUN mkdir -p /usr/src/app
-WORKDIR /usr/src/app
+RUN useradd --user-group --create-home --shell /bin/false app
 
-COPY package.json /usr/src/app/
+ENV HOME=/home/app
+
+COPY package.json $HOME
+RUN chown -R app:app $HOME/*
+
+USER app
+WORKDIR $HOME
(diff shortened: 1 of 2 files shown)


beenanner:
I wonder if we could run this as nobody. Will test both out later tonight.

beenanner:
Works on Ubuntu with "app" user. Just a note using "nobody" didn't work since we expose the volume for mounting to the host machine during `docker run` and it overrides the folder permissions for /sitespeed.io to be the first non-root user(uid 1000).

```
➜  ~ docker exec -it ddf8 bash 
app@ddf8ba49b527:/sitespeed.io$ whoami
app
app@ddf8ba49b527:/sitespeed.io$ top
 PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND                                            
  180 app       20   0  534072  15080   1392 S   0.3  0.2   0:00.07 chrome                                             
    1 app       20   0 1196764 146496  18820 S   0.0  1.8   0:02.64 node                                               
   20 app       20   0  169976  24088  14176 S   0.0  0.3   0:00.06 Xvfb      
```

Issue #1267, opened 2016-10-25
Wrong exit status when testing URLs that do not work


soulgalore:
<pre>
$ bin/sitespeed.js https://failing -n 1
...
$ echo $?
0
</pre>

beenanner:
Looks like this is caused by https://github.com/sitespeedio/sitespeed.io/blob/master/bin/sitespeed.js#L39

soulgalore:
good, that is one thing: that check should only run when we configured a budget, I'll fix that first.

soulgalore:
I think we can just check for errors in the budget, I've missed that when I implemented it.

soulgalore:
no that will not work :) lets think about it for a while.

beenanner:
We possible might need something upstream returned in browsertime for the status of the initial URL.

soulgalore:
Yes. I think there's an issue to add an error element in the JSON from Browsertime, but it will be some work so it has been put aside. One ugly solution for now could be to compare number of URLs in the budget and the ones tested or something like that.

jchoksi-whitbread:
Do you think it would be possible to somehow check for the test URL failing to load in the --browsertime.preScript and do something in the script to stop the test? If so, what could we do to stop the sitespeed.io process from within the preScript?

I'm basically looking for a way to get notified when a Jenkins job running sitespeed.io against a URL fails because the target site is not up. Currently, as per above, the exit code returned is still 0 for v5.6.5:

```
[2017-11-01 18:33:28] INFO: Versions OS: linux 4.9.0-4-amd64 nodejs: v6.11.5 sitespeed.io: 5.6.5 browsertime: 1.9.4 coach: 0.36.0
[2017-11-01 18:33:28] INFO: Starting chrome for analysing https://failing 1 time(s)
[2017-11-01 18:33:28] INFO: Testing url https://failing run 1
[2017-11-01 18:33:32] ERROR: Could not load URLUrlLoadError: Failed to load https://failing
[2017-11-01 18:33:32] ERROR: https://failing generated the following error in Browsertime UrlLoadError: Failed to load https://failing
    at BrowsertimeError (/opt/npm-global/lib/node_modules/sitespeed.io/node_modules/browsertime/lib/support/errors.js:5:5)
    at UrlLoadError (/opt/npm-global/lib/node_modules/sitespeed.io/node_modules/browsertime/lib/support/errors.js:19:5)
[...]

PR #1268, opened 2016-10-25
Temp fix for running against a bad URL #1267

beenanner:
This is just a temp fix as the HTML output is still rendered, but at least the exit code is correct. :-p


beenanner:
Nvm this won't work.

PR #1279, opened 2016-10-29
Fix header level for 4.0.0 release

svetlyak40wt:
This way it will not break automatic parsers like https://allmychanges.com/p/javascript/sitespeed.io/


soulgalore:
Thank you @svetlyak40wt !

PR #1286, opened 2016-10-31
Catching exit code from browsertime and exiting non-zero is returned

beenanner:


lib/plugins/browsertime/analyzer.js
```diff
@@ -73,7 +73,11 @@ module.exports = {
     let engine = new browsertime.Engine(btOptions);
     log.info('Starting %s for analysing %s %s time(s)', btOptions.browser, url, btOptions.iterations);
     return engine.start()
-      .then(() => engine.run(url, scriptsByCategory))
+      .then((result,error) => engine.run(url, scriptsByCategory))
+      .catch((error) => {
+        process.exitCode = 1;
+        process.exit();
```
beenanner: Ah I see what you mean let's focus on fixing it in browsertime via a result json then. :-)

soulgalore: ah sorry I should have been clearer with that, I'll make that issue cleaner.

beenanner: nah no worries only spent a couple of minutes on that change :-p

Issue #1068, opened 2016-07-22
Let cli help show help from browsertime


soulgalore:
Is there's a way to let --help show the help instructions from browsertime and add browsertime. before each parameter?

soulgalore:
We have more important things to focus on.

PR #1295, opened 2016-11-01
Browsertime options 

beenanner:
Created this branch to resolve conflicts and merge in changes from #1068. Everything appears to be working @tobli / @soulgalore. We should be able to merge in https://github.com/sitespeedio/browsertime/pull/201 tag it then we can update this pull-request to pull in that tag instead of the the branch.

### DIFF
--- a/bin/sitespeed.js
+++ b/bin/sitespeed.js
@@ -7,7 +7,6 @@
 const cli = require('../lib/support/cli'),
   sitespeed = require('../lib/sitespeed'),
   Promise = require('bluebird'),
-  loader = require('../lib/support/pluginLoader'),
   browsertimeCli = require('browsertime').cli;
 
 if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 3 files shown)


soulgalore:
This looks cool, haven't checked the code but we don't support the validateInput function in Browsertime? I think we should think about a pattern for that, then we are ready to go +  we can then implement the same for all plugins, so that if you add your own plugin, you can get the CLI options from help, that will be super cool!

beenanner:
What if for the first iteration we add an array of which options to delete to the opt object in sitespeed.io that can then be used by the plugin to remove them. Iterating over the options deleting them before the prefixing/grouping of the plugin options. 

Ideally it would be nice to set it up where you can just drop a plugin in and everything just works or  even create a pattern where someone could just `npm install sitespeedio-plugin-<plugin-name> --save` from within sitespeedio and sitespeedio would able to auto discover and load them.

soulgalore:
Closed the corresponding ones in Browsertime, lets do the same here, focus on what's important.

PR #1361, opened 2016-12-03
WIP: Upgrade the InfluxDB module.

tobli:
I'm pushing this completely untested. Would love for someone to give it a spin and see if it actually works. Version 5 of the influx module has a completely new API.


eripa:
I've given this a spin using the [docker compose approach](https://raw.githubusercontent.com/sitespeedio/sitespeed.io/master/docker/docker-compose.yml), but replacing the graphite container with a custom built container from this commit. 

`docker run --rm --net=container:sitespeed_influxdb_1 sitespeed-upgrade-influx http://developer.sony.com  -b firefox  --plugins.load influxdb --influxdb.host influxdb --influxdb.database sitespeed_io`

It works fine as far as I can see. I'm able send data with the below caveat. I think they are more generic issues for InfluxDB support than issues with this upgrade. Should I add them to #889?

- I need to manually create the InfluxDB database before starting to send data
- each metric is sent as a complete series, which means that I cannot break it up on various levels (like it's done using Graphite). This is the same behavior as I see when I run InfluxDB with a Graphite input plugin. I'm not sure of this, but it looks like the point should be sent with a comma instead of a period. see [Writing data](https://docs.influxdata.com/influxdb/v1.1/guides/writing_data/)

tobli:
Thanks for the testing @eripa. Given that it seems to work I'll rebase, uncomment the cli documentation, and merge this. With this we should be able to close #889. It'd be lovely if you could file your two suggestions above as two separate issues. Thanks for the feedback!

soulgalore:
> each metric is sent as a complete series, which means that I cannot break it up on various levels (like it's done using Graphite)

hmm shouldn't we see if we can fix this before we close #889 ?

eripa:
Personally I would recommend looking at fixing the metric split issue. Having it the way it is virtually removes the possibility to use variables in dashboards as you will need to create a variable with the complete metric series name.

soulgalore:
Also in the compose file the database name is sitespeed and in your example run @eripa it's sitespeed_io, could that be the problem?

eripa:
The problem in what issue, @soulgalore? The fact that I needed to create the database manually?

Or do you think it will make any difference having the Influx database be named just sitespeed in regards to the metric split issue?

soulgalore:
ah sorry @eripa yes I meant that in your example the database was called sitespeed_io but in the compose file it is sitespeed only, but maybe did your own setup? Maybe it's also better to have a create script instead of specific tutum/influxdb thing. however the other issue seems more important.

eripa:
Ah, I see.  No worries! Yes I used my own docker-compose file.. And in the container image I uses there doesn't seem to be any "precreate" functionality. 

Excerpt for my docker-compose:
```
  influxdb:
    image: influxdb:alpine
    volumes:
      - ./influxdb-data:/var/lib/influxdb/data
      - ./influxdb-wal:/var/lib/influxdb/wal
      - ./influxdb.conf:/etc/influxdb/influxdb.conf:ro
    restart: always
    command: influxd -config /etc/influxdb/influxdb.conf
```

soulgalore:
Cool, I see now, we should also switch to the official one (and bind to a specific version). About the database: I guess we can use the API as in https://hub.docker.com/_/influxdb/ 

Also think we should have some default dashboard when we release. We can merge the PR but I think we should have the cli options commented out for now until we have everything fixed.

eripa:
Sounds good to me! Should I create a separate issue for the split metric issue?

soulgalore:
@eripa yes please and examples if you have :)

aliradd:
Hi there,
If you guys like someone to load test the write capability I have a solution ready to go.
I would need a little help to get this plugging up and running but on the load generation side my tool is ready. We can perhaps try to see how much we could push into influx etc....not sure if this is of interest to you?

PR #1372, opened 2016-12-10
Use video.js as videoplayer

soulgalore:
We get a better look and feel using video.js and +200kb JS :(. But that will be less when they skip support for flash and that will happen soonish.


soulgalore:
@beenanner yes I'll fix the marging, good. About the mobile, seen that too but not 100% best way to fix that.

PR #1379, opened 2016-12-12
Trap exit, inspired by docker selenium

soulgalore:


tobli:
+1
Should have the same for Browsertime as well!

PR #1417, opened 2017-01-17
usage of --shm-size requires at least docker 1.10

staabm:
the --shm-size option was added with docker 1.10. mention the min docker version in the readme.

see https://github.com/docker/docker/pull/16168


soulgalore:
thanks @staabm !

Issue #418, opened 2014-05-27
Implement check for how much CSS that is used (and use it for critical path score)


soulgalore:
https://github.com/giakki/uncss

soulgalore:
Closing too old issues. Lets start with new clean sheet for 4.0.

PR #1421, opened 2017-01-18
Define budget per run #418

soulgalore:
There could be other things that we need to fix to be able to re-run the same object.


soulgalore:
No real need to fix this, lets close it.

Issue #1422, opened 2017-01-23
Possibility to also pass wpt.script to sitespeed.io through string instead of file


JeroenVdb:
Hi all!

I'm setting up some sitespeed.io, WPT, Graphite/Grafana monitor in AWS.
Sitespeed.io will be the driver for running test (in WPT) and sending data to Graphite/Grafana setup.

I would like to create an HTTP API endpoint + lambdaFunction that starts an sitespeed.io Docker (ECS) instance so teams can easily run en schedule monitor tests.

For this they would send all config (mainly sitespeed.io config) in the body of the request. 
Because some config options (urls, budget and webpagetest script) expect a path and file this isn't really possible.

**Would it be a good idea to create an extra config where you can send a webpagetest script that is already converted to a string? For this I would use the WebPageTest API method scriptToString().
Or do you guys see other options?**

https://github.com/marcelduran/webpagetest-api/#methods

soulgalore:
Hey @JeroenVdb let us fix so that we check if it is a file or a text, I can do that the coming days. Then you can automagically pass on a file or the script.
Best
Peter

JeroenVdb:
Did some testing. The cli command would be something like this:

```
# via direct script string
sitespeed.io --webpagetest.script "logData\t0\nnavigate\thttp://foo.com/login\nlogData\t1\nsetValue\tname=username\tjohndoe\nsetValue\tname=password\t12345\nsubmitForm\taction=http://foo.com/main" http://sitespeed.io

# via file
sitespeed.io --webpagetest.script "./plain-wpt-script.txt" http://sitespeed.io
```

```javascript
$ node
> var fs = require('fs');
> var webpagetest = require('webpagetest');
> 
> // array to webpagetest script string
> wptScriptString = webpagetest.scriptToString([
... 	{"logData": 0},
... 	{"navigate": "http://foo.com/login"},
... 	{"logData": 1},
... 	{"setValue": ["name=username", "johndoe"]},
... 	{"setValue": ["name=password", "12345"]},
... 	{"submitForm": "action=http://foo.com/main"}
... ]);
'logData\t0\nnavigate\thttp://foo.com/login\nlogData\t1\nsetValue\tname=username\tjohndoe\nsetValue\tname=password\t12345\nsubmitForm\taction=http://foo.com/main'
> 
> /*
... 	plain-wpt-script content
... 
... 	logData	0
... 	navigate	http://foo.com/login
... 	logData	1
[...]

Issue #1399, opened 2016-12-27
JUnit / TAP reporting is not working


dhanabalanvk:
Hi Team,

I am trying to integrate JUnit / TAP report functionality with Jenkin but report files are not generated while running the below command,

$ sitespeed.io https://www.sitespeed.io/ --budget myBudget.json --budget.output junit -b chrome -n 5

**As per the documentation it should create a junit.xml in the outputFolder.**

$ sitespeed.io https://www.sitespeed.io/ --budget myBudget.json --budget.output tap -b chrome -n 5

**As per the documentation it should create a budget.tap in the outputFolder.**

Could you please assist me whether I am missing anything to generate the reports?

Regards,
Dhanabalan Venkitusamy

soulgalore:
Hey @dhanabalanvk thanks for reporting, this is bug (like it is now the the budget file overwrites the budget config) that I introduced a while ago, let me see if I can fix that tomorrow.

Best
Peter

soulgalore:
I tracked it back to an upgrade of yargs. If I revert back to 5.0 it works, when we upgraded 6.x it failed.

soulgalore:
Hey @dhanabalanvk I pushed 4.3.1 now where it works. It was a bad upgrade of yargs that caused the problem, I rolled back and will do a better fix after the holidays.

soulgalore:
this is released and fixed.

PR #1431, opened 2017-01-26
#1422 Make it possible to insert a direct string with a wpt script into webpagetest.script

JeroenVdb:
This enables using a direct WebPageTest script string as the input for the `--webpagetest.script`  flag.

Example:
```
// NEW
./bin/sitespeed.js --webpagetest.key <key> --webpagetest.script "logData\t0\nnavigate\thttp://www.jeroenvdb.be\nlogData\t1\nnavigate\thttp://www.demorgen.be" http://www.jeroenvdb.be

// NEW
./bin/sitespeed.js --webpagetest.key <key> --webpagetest.script  "logData\t0\nnavigate\thttp://www.demorgen.be\nlogData\t1\nnavigate\t{{{URL}}}" http://www.jeroenvdb.be

// still possible to use a script from path
./bin/sitespeed.js --webpagetest.key <key> --webpagetest.script ./wpt-script.txt
```

### DIFF
--- a/lib/support/cli.js
+++ b/lib/support/cli.js
@@ -283,7 +283,7 @@ module.exports.parseCommandLine = function parseCommandLine() {
       group: 'WebPageTest'
     })
     .option('webpagetest.script', {
-      describe: 'Path to a script file',
+      describe: 'Path to a script file or an already parsed string with the script',
       group: 'WebPageTest'
     })
     /** Google Page Speed Insights */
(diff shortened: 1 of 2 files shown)


soulgalore:
Hey @JeroenVdb ah sorry, we cannot upgrade yargs to 6 because there was a breaking change between 5 -> 6 and we haven't fixed that yet, see https://github.com/sitespeedio/sitespeed.io/issues/1399#issuecomment-269437628

PR #1434, opened 2017-01-29
Send annotations to Graphite for each tested URL

soulgalore:
This is an early version of sending annotations to Graphite for each tested URL. The annotation event will have tags that match the templates in pageSummary, so you easy can pick up the exact event.

To test it out on your local, start graphite/grafana with our docker-compose file and then run:
<pre>
bin/sitespeed.js http://www.sitespeed.io -n 1 --graphite.host 127.0.0.1 --graphite.auth LOGIN:PASSWORD -b chrome --resultBaseURL https://results.sitespeed.io
</pre>

Then in Grafana add your annotation (more or less $domain $page $browser $connectivity):
screen shot 2017-01-29 at 13 38 29

This in on the pageSummary dashboard 4.0 (for 4.1 you can also add the base variable). Use the latest tag for the grafana bootstrap and it should work out of the box.

We need to try this out ... one way when we feel the code is ready is to switch on dashboard.sitespeed.io for a day or two to run latest docker before we release.

The result:
screen shot 2017-01-29 at 13 58 04


soulgalore:
@tobli @beenanner anyone of you volunteering to have a look? wanna push this the coming weekend and love your feedback :)

beenanner:
@soulgalore tested and LGTM. There is a caveat users would need to be aware of if they use templatized variables with annotations. Seems the annotations only use the variable on dashboard load if you are using the multi select / all option. It does work fine toggling if you only allow one value per variable though so good enough for me. We can probably just document the caveat with multi-value in the docs for users.

lib/plugins/graphite/send-annotation.js
```diff
@@ -0,0 +1,62 @@
...
+    // The tags make it possible for the dashboard to use the
+    // templates to choose which annotations that will be showed.
+    // That's why we need to send tags that matches the template
+    // variables in Grafana.
+    const connectivity = graphiteUtil.getConnectivity(options);
+    const browser = options.browser;
+    const namespace = options.graphite.namespace.split('.').join(',');
+    const urlAndGroup = graphiteUtil.getURLAndGroup(options, group, url, options.graphite.includeQueryParams).split('.').join(',');
+    const tags =  `${connectivity},${browser},${namespace},${urlAndGroup}`;
+    const message = `<a href='${options.resultBaseURL}/${baseDir}/${pagePath}'>Result ${options.browsertime.iterations} runs</a>`;
```
soulgalore: yes good, let me fix that!

soulgalore:
@beenanner are you sure about:
>  Seems the annotations only use the variable on dashboard load if you are using the multi select / all option

Not 100 sure I fully understand, I need to test it some more then. When I tested before the annotation set on https://dashboard.sitespeed.io/dashboard/db/page-summary worked for me (no multi/all options) but I only tested it locally. Maybe there's an Grafana issue for it.

beenanner:
Oops @soulgalore that was meant to be if you're **NOT** using the multi select / all option. (-:

soulgalore:
ah ok :)

PR #1441, opened 2017-02-02
Allow budget to be specified inline on cli.

JeroenVdb:
I've updates the yargs (cli options library) to version 6.
We reverted this update in #1399 because if caused a conflict between the --budget and --budget.output flags (was automagically fixed in 5 but not anymore in 6).

Using the new version seemed logical for #1431 and because we don't want to be pinned down on an old version of yargs.

### DIFF
--- a/lib/support/cli.js
+++ b/lib/support/cli.js
@@ -190,8 +190,12 @@ module.exports.parseCommandLine = function parseCommandLine() {
       group: 'Plugins'
     })
     /** Budget */
-    .option('budget.file', {
+    .option('budget.configPath', {
       describe: 'Path to the JSON budget file.',
+      group: 'Budget'
+    })
+    .option('budget.config', {
+      describe: 'The JSON budget config as a string.',
       group: 'Budget'
     })
     .option('budget.output', {
@@ -394,10 +398,13 @@ module.exports.parseCommandLine = function parseCommandLine() {
     .alias('help', 'h')
     .config('config')
     .alias('version', 'V')
+    .conflicts('budget.config', 'budget.configPath')
     .coerce('budget', function(arg) {
-      if (arg === Object(arg) && !Array.isArray(arg)) {
-        if (arg.hasOwnProperty('file')) {
-          arg.file = JSON.parse(fs.readFileSync(arg.file, 'utf8'))
+      if (typeof arg === 'object' && !Array.isArray(arg)) {
+        if (arg.configPath) {
+          arg.config = JSON.parse(fs.readFileSync(arg.configPath, 'utf8'));
+        } else if (arg.config){
+          arg.config = JSON.parse(arg.config);
         }
         return arg;
       } else {
--- a/lib/support/cli.js
+++ b/lib/support/cli.js
@@ -190,7 +190,7 @@ module.exports.parseCommandLine = function parseCommandLine() {
       group: 'Plugins'
     })
     /** Budget */
-    .option('budget', {
+    .option('budget.file', {
       describe: 'Path to the JSON budget file.',
       group: 'Budget'
     })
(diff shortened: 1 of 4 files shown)


lib/support/cli.js
```diff
@@ -395,7 +395,14 @@ module.exports.parseCommandLine = function parseCommandLine() {
     .config('config')
     .alias('version', 'V')
     .coerce('budget', function(arg) {
-      return JSON.parse(fs.readFileSync(arg, 'utf8'))
+      if (arg === Object(arg) && !Array.isArray(arg)) {
```
tobli: Object(arg) looks a bit strange to me, maybe go with typeof arg === 'object'.

lib/support/cli.js
```diff
@@ -395,7 +395,14 @@ module.exports.parseCommandLine = function parseCommandLine() {
...
+        if (arg.hasOwnProperty('file')) {
```
tobli: if (arg.file) should be fine here I think.

package.json
```diff
@@ -97,6 +97,6 @@
     "uuid": "^3.0.0",
     "webcoach": "0.30.4",
     "webpagetest": "0.3.4",
-    "yargs": "5.0.0"
+    "yargs": "6.4.0"
```
tobli: 6.6.0 is the most recent now.

tobli:
The coerce is great, however I'd like to do some other cleanup to budget/index.js at the same time. Budget.file is a required property, but we don't check for it. Misspelling a cli property (e.g. --budget.flie=foo) results in a crash. The fix is to check for required properties in open:
```
    throwIfMissing(options.budget, ['file'], 'budget');
```

With that the if statements in processMessage and close can be removed, and we avoid the crash. I'd also recommend pulling out the relevant properties from the options in open, i.e.
```
    this.budget = options.budget.file;
    this.output = options.budget.output;
```

I hope it's ok to make these changes.

tobli:
Building on this I propose to not overload parameters to be both path and data (for budget and in #1431). I think a better pattern is ```--foo.file```and ```--foo.data, --foo.script, --foo.json``` or whatever fits best. That way the coerce could pick up .file, read the data and fill in the .data property. That retains the original value, reads more natural when the property is read, and works better if the plugin is used standalone. Given that, maybe ```path``` is a better name (e.g. ```--budget.path```), if we want to set a standard. Thoughts? ping @soulgalore @beenanner

JeroenVdb:
Thanks for the input, I will update my PR.

Ironically I discovered another strange new behaviour or bug when using the new Yargs version, I created a ticket: https://github.com/yargs/yargs/issues/777

tobli:
Oh, that's interesting. Great bug report, let's hope for a quick resolution

JeroenVdb:
@tobli I would rather not cleanup the budget/index.js in **this** PR but create another one for that issue/improvement to keep the PRs clean and separated by concerns. Is this ok?

I totally agree to use more clear naming. Maybe configPath and scriptPath is more clear?
```
--budget.configPath and --budget.config (can be directly inserted or via budget.configPath)
--webpagetest.scriptPath and --webpagetest.script (can be directly inserted or via webpagetest.scriptPath)

// or
--budget.path and --budget.config
--webpagetest.path and --webpagetest.script
```

tobli:
Absolutely, if you could split the PRs that would be awesome! I really like the naming suggestion, I think that could work as a general rule.

JeroenVdb:
@tobli I updated the PR with your comments.

```
./bin/sitespeed.js https://www.sitespeed.io/ --budget.config {"coach.pageSummary": [{"metric": "advice.info.domDepth.max","max": 10}]} --budget.output junit -b chrome -n 1
./bin/sitespeed.js https://www.sitespeed.io/ --budget.config "{\"coach.pageSummary\": [{\"metric\": \"advice.info.domDepth.max\",\"max\": 10}]}" --budget.output junit -b chrome -n 1
```

```
{
  "coach.pageSummary": [{
    "metric": "advice.info.domDepth.max",
    "max": 10
  }]
}
```

tobli:
I updated the title to better reflect the change (the yargs upgrade is not the central part).

tobli:
Awesome change, thanks!

beenanner:
My preference would be `--budget.config` and `--budget.file` as a standard.

Issue #1440, opened 2017-02-01
Investigate best practice for plugin support for Docker images


beenanner:
Will need to update the current docs with best practices. In short use volumes and if you need to share an image you can bake them into the image directory and copy them to your image during build time

PR #1442, opened 2017-02-03
Adding docs around plugins and docker #1440

beenanner:
Before merging I think it's probably worth creating a simple hello-world plugin repo as a way to test that this continues to work as well as it will give a concrete example for users to follow. Thoughts?


beenanner:
Actually this is probably ok as is if we want to create a example plugin that can be done later. :-)

PR #1463, opened 2017-02-19
Added Firefox & Chrome specific options

soulgalore:
Is there a better way to group the options, check https://github.com/yargs/yargs#groupkeys-groupname


tobli:
I'm not sure it's worthwhile to have separate groups from Browser, just for one or two options.

soulgalore:
One/two today, 10000 tomorrow :) Lets push it in browser then.

beenanner:
Yeah agreed keeping them in browser group for now seems best

PR #1465, opened 2017-02-19
Mistake in the documentation.

davidbonnici1984:
Mistake in the documentation.


soulgalore:
Hey @davidbonnici1984 thanks for sending the PR! Actually, it should be both port and httpPort:The port is for sending metrics and the httpPort is for events and user interface. I'll update the docs!

Best
Peter

Issue #1475, opened 2017-02-26
Compress traceCategories timeline JSON from Chrome by default


soulgalore:
The timeline JSON can be massive, we should think about how we can compress it by default.

soulgalore:
released

PR #1479, opened 2017-02-28
Compress the trace log #1475

soulgalore:


lib/plugins/tracestorer/index.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
       case 'browsertime.chrometrace':
       case 'webpagetest.chrometrace':
         {
-          this.storageManager.writeDataForUrl(JSON.stringify(message.data, null, 0), message.name, message.url);
```
tobli: Good catch! I'd expect Bluebird to warn about this. Maybe we didn't run it enough in non-prod mode to notice.

lib/support/storageManager.js
```diff
@@ -6,16 +6,29 @@ const fs = require('fs-extra'),
...
   return Promise.join(dirPath, filename, data,
-    (dirPath, filename, data) =>
-      fs.writeFileAsync(path.join(dirPath, filename), data, 'utf8'));
+    (dirPath, filename, data) => {
+      if (gzip) {
+        const buff = new Buffer(data, 'utf8');
+        return zlib.gzipAsync(buff, {level: 1}).then((buffer) =>
+          fs.writeFileAsync(path.join(dirPath, filename + '.gz'), buffer, 'utf8')
+        );
+      }
```
tobli: Just a nitpick, but this isn't how we typically place the braces for else, is it?

soulgalore: I'll fix that!

lib/plugins/tracestorer/index.js
soulgalore: it warn in some cases where it is obvious I think.

tobli: It's supposed to warn if a Bluebird promise was created, but not returned. Anyways, good that this is fixed.

http://bluebirdjs.com/docs/warning-explanations.html#warning-a-promise-was-created-in-a-handler-but-was-not-returned-from-it

PR #1507, opened 2017-03-11
Refactor Slack plugin for readability.

tobli:
* Use lodash.get to simplify optional properties
* Sync metric titles for summary and attachments.

### DIFF
--- a/lib/plugins/slack/attachements.js
+++ b/lib/plugins/slack/attachements.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const get = require('lodash.get');
 const h = require('../../support/helpers');
 
 function getMetric(metric) {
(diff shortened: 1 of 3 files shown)


lib/plugins/slack/attachements.js
```diff
@@ -12,89 +13,100 @@ function getMetric(metric) {
...
             short: false
           })
         }
       }
 
-      let color = metrics[options.limitMetric] ? metrics[options.limitMetric].metric > options.limitError ? 'good' : metrics[options.limitMetric].metric > options.limitWarning ? 'warning' : 'danger' : 'danger';
+      let color = 'danger';
 
-      attachements.push({
+      const limitMetric = metrics[options.limitMetric];
```
tobli: This looks strange to me. Shouldn't 'good' be the default? Also, the other checks seems backwards. For coachScore, higher is better, but for speedIndex and firstVisualChange the reverse is true.

lib/plugins/slack/summary.js
```diff
@@ -1,75 +1,74 @@
...
+      name: 'Fully Loaded',
+      metric: get(base.browsertime, 'summary.fullyLoaded.median')
+    },
+    coachScore: {
+      name: 'Coach score',
+      metric: get(base.coach, 'summary.performance.score.median')
+    },
+    transferSize: {
+      name: 'Page transfer weight',
+      metric: h.size.format(get(base.pagexray, 'summary.transferSize.median'))
```
tobli: This is a slight change that might not be desirable, formatting a missing transfer size as N/A.

tobli:
Some small tweaks to reduce long lines and complex statements.

lib/plugins/slack/attachements.js
soulgalore: yep this should be changed.

lib/plugins/slack/summary.js
soulgalore: I think that's ok for now

tobli:
Cool. Just some drive-by changes to make it easier to make tweaks later.

Issue #1508, opened 2017-03-11
[Slack] SpeedIndex/firstVisualChange check is wrong


soulgalore:
https://github.com/sitespeedio/sitespeed.io/pull/1507#discussion_r105529708

tobli:
This is fixed in Sitespeed.io 4.7.0

PR #1509, opened 2017-03-11
Timing metrics should be lower than to get a green light #1508

soulgalore:
The naming error/warning could be better, but push that to 5.0 👯‍♂️


lib/plugins/slack/attachements.js
```diff
@@ -93,10 +93,20 @@ module.exports = {
       let color = 'danger';
```
tobli: This should be 'good', in case the metric isn't available (e.g. default of coachScore when not running coach)

lib/plugins/slack/attachements.js
```diff
@@ -93,10 +93,20 @@ module.exports = {
...
 
       const limitMetric = metrics[options.limitMetric];
-      if (limitMetric) {
+      if (limitMetric && options.limitMetric === 'coach') {
```
tobli: The metric is 'coachScore'

tobli: Also the default breakpoints for coachScore are inverted, error should be lower than warning.

tobli: This check would be:
<= error: 'danger'
else if
<= warning: 'warning'
(no else needed if 'good' is default)

lib/plugins/slack/attachements.js
```diff
@@ -93,10 +93,20 @@ module.exports = {
...
           color = 'good';
-        } else if (limitMetric.metric > options.limitWarning) {
+        }
+        else if (limitMetric.metric > options.limitWarning) {
+          color = 'warning';
+        }
+      }
+      // SpeeddIndex/firstVisualChange
+      else if (limitMetric) {
+        if (limitMetric.metric < options.limitError) {
```
tobli: here warning would be lower than error (we might want to verify that). Hence:
>= error: 'danger'
else if >= warning: 'warning'
(default to 'good')

tobli: Hmm, accidental markdown... should be >= error

soulgalore: yes but the default values in the cli is kind of a mess (error higher than warning for coach score), that's my comment about 5.0 or we just fix that now.

tobli:
👍🏻

PR #1515, opened 2017-03-14
gzip HAR files

soulgalore:


lib/plugins/html/templates/url/waterfall/fetchHAR.pug
```diff
@@ -15,11 +16,27 @@ script(type='text/javascript').
...
-  if (window["fetch"]) {
-      window["fetch"]('data/browsertime.har')
-      .then(f => f.json().then(j => {
-        var har = j.log;
-        var perfCascadeSvg = perfCascade.fromHar(har, options);
-        outputHolderEl.appendChild(perfCascadeSvg);
-      }))
-  }
+if options.gzipHAR
+    script(type='text/javascript').
```
tobli: Could we put the gunzip.min.js script tag here? To only include it when needed.

lib/plugins/html/templates/url/waterfall/fetchHAR.pug
```diff
@@ -15,11 +16,27 @@ script(type='text/javascript').
...
+      if (window["fetch"]) {
+          window["fetch"]('data/browsertime.har.gz')
+          .then(f => f.arrayBuffer().then(arrayBuffer => {
+            var byteArray = new Uint8Array(arrayBuffer);
+            var gunzip = new Zlib.Gunzip(byteArray);
+            var decompressedArray = gunzip.decompress();
+            var string = new TextDecoder("utf-8").decode(decompressedArray);
+            var har = (JSON.parse(string)).log;
```
tobli: I prefer to pass the full HAR (not just the log part) to PerfCascade.

lib/plugins/html/templates/url/waterfall/fetchHAR.pug
```diff
@@ -15,11 +16,27 @@ script(type='text/javascript').
...
+            var perfCascadeSvg = perfCascade.fromHar(har, options);
+            outputHolderEl.appendChild(perfCascadeSvg);
+          }))
+      }
+else
+    script(type='text/javascript').
+      if (window["fetch"]) {
+       window["fetch"]('data/browsertime.har')
+       .then(f => f.json().then(j => {
+         var har = j.log;
```
tobli: Same as above. Skip this line, and rename j to har.

lib/plugins/html/templates/url/waterfall/fetchHAR.pug
```diff
@@ -15,11 +16,27 @@ script(type='text/javascript').
...
+            var har = (JSON.parse(string)).log;
+            var perfCascadeSvg = perfCascade.fromHar(har, options);
+            outputHolderEl.appendChild(perfCascadeSvg);
+          }))
+      }
+else
+    script(type='text/javascript').
+      if (window["fetch"]) {
+       window["fetch"]('data/browsertime.har')
+       .then(f => f.json().then(j => {
```
tobli: Change the parenthesis (adding one here, remove one at the end), to avoid unneeded nesting of the promise flow. Use
```
...
.then(f => f.json())
.then(har => { ...})
```

lib/support/cli.js
```diff
@@ -447,6 +447,11 @@ module.exports.parseCommandLine = function parseCommandLine() {
     .option('resultBaseURL', {
       describe: 'The base URL to the server serving the HTML result. In the format of https://result.sitespeed.io'
     })
+    .option('gzipHAR', {
+      describe: 'Compress the HAR files with GZIP.',
+      default: false,
```
tobli: This is so awesome! Why not turn it on by default?

lib/plugins/html/templates/url/waterfall/fetchHAR.pug
```diff
@@ -15,11 +16,27 @@ script(type='text/javascript').
...
-        var har = j.log;
-        var perfCascadeSvg = perfCascade.fromHar(har, options);
-        outputHolderEl.appendChild(perfCascadeSvg);
-      }))
-  }
+if options.gzipHAR
+    script(type='text/javascript').
+      if (window["fetch"]) {
+          window["fetch"]('data/browsertime.har.gz')
+          .then(f => f.arrayBuffer().then(arrayBuffer => {
```
tobli: Same as below, change the promise chain to
```
.then(f => f.arrayBuffer())
.then(arrayBuffer => ...)

lib/support/cli.js
soulgalore: We will change the current behavior if we have it on by default. It adds one extra step if you download and wanna drag/drop it to another tool, like the maybe coming "compare 2 HARs" :)

lib/plugins/html/templates/url/waterfall/fetchHAR.pug
soulgalore: yes

lib/support/cli.js
tobli: True… =(
But we should follow up with gzip support in PerfCascade! And maybe offer "on the fly" decompression if possible (Download as HAR or Download as HAR.gz). Shape the future! =)

tobli: So false for now is ok.

lib/plugins/html/templates/url/waterfall/fetchHAR.pug
soulgalore: Yep, we should change the example in the README in PerfCascade. I'll change here.

lib/plugins/html/templates/url/waterfall/fetchHAR.pug
soulgalore: cool will fix

lib/support/cli.js
soulgalore: yep lets do it like that for now

Issue #1519, opened 2017-03-16
Add link to the result in Slack


soulgalore:
If you configured <code>resultBaseURL</code> we should add the link to the HTML result page in the Slack message, so you easily can access it.

soulgalore:
fixed.

PR #1523, opened 2017-03-16
Added links to result page for Slack #1519

soulgalore:


lib/plugins/slack/attachements.js
```diff
@@ -11,7 +11,7 @@ function getMetric(metric) {
   }
 }
 
-module.exports = function(dataCollection, slackOptions) {
+module.exports = function(dataCollection, slackOptions, resultBaseURL, storageManager) {
```
tobli: I think it'd be good to place the slackOptions map last (that's the pattern we use for other functions that take options).

lib/plugins/slack/summary.js
```diff
@@ -3,7 +3,7 @@ const util = require('util');
 const get = require('lodash.get');
 const h = require('../../support/helpers');
 
-module.exports = function(dataCollection, errors, options) {
+module.exports = function(dataCollection, errors, options, storageManager) {
```
tobli: options last, as above.

lib/plugins/slack/attachements.js
```diff
@@ -108,9 +108,16 @@ module.exports = function(dataCollection, slackOptions) {
...
+    if (resultBaseURL) {
+      const baseDir = storageManager.getRelativeBaseDir();
+      const pagePath = storageManager.pathFromRootToPageDir(url);
+      message = ` (<${resultBaseURL}/${baseDir}/${pagePath}|result>)`;
+    }
+
     attachments.push({
       color,
-      text: url,
+      text: url + message,
```
tobli: Maybe if the message variable was called text, initiated to url and appended to inside the if-statement. Then the attachment object could consistently use shorthand syntax.

lib/plugins/slack/attachements.js
```diff
@@ -108,9 +108,16 @@ module.exports = function(dataCollection, slackOptions) {
       }
     }
 
+    let message = '';
+    if (resultBaseURL) {
+      const baseDir = storageManager.getRelativeBaseDir();
```
tobli: I think we should reconsider the approach with relativeBaseDir. It's not a good fit for use with outputFolder. ```--resultBaseURL="http://example.com/bar" --outputFolder="../../foo"``` will result in ```http://example.com/foo/...```when formatted by Slack. ```--resultBaseURL="http://example.com/bar" --outputFolder="/var/www/foo"``` becomes ```http://example.com/bar//var/www/foo``` which is probably not what the user wanted.

soulgalore: We have https://github.com/sitespeedio/sitespeed.io/blob/720d3b93c21034ce4ded3707916276a92888fb1a/lib/plugins/s3/index.js#L39 that makes the combo of S3 and outputFolder a combo that doesn't work. I think we should create another issue to fix this longterm

tobli: Ahh, I see. Sounds like a plan. Merge this, create a follow up issue.

soulgalore: https://github.com/sitespeedio/sitespeed.io/issues/1527 let us discuss it there!

PR #1537, opened 2017-03-21
Minor cleanup and tweaks to S3 plugin.

tobli:
* Avoid hang if fs.remove fails.
* Use defaults from s3 module for maxAsyncS3, s3RetryCount etc.
* Always set whichever of region, accessKeyId and secretAccessKey that are passed as options. That allows for an arbitrary mix of options and env. variables.
* Allow configuration of  S3 (such as s3RetryDelay) even when credentials aren’t passed as options.
* Use verbose logging instead of trace (since trace prints a stack trace).

### DIFF
--- a/lib/plugins/s3/index.js
+++ b/lib/plugins/s3/index.js
@@ -5,78 +5,75 @@ const s3 = require('s3');
 const Promise = require('bluebird');
 const throwIfMissing = require('../../support/util').throwIfMissing;
 const fs = require('fs-extra');
+const pick = require('lodash.pick');
 
 Promise.promisifyAll(fs);
 
-function getClient(options) {
-  if (options.s3.key && options.s3.secret) {
-    const settings = {
-      maxAsyncS3: options.s3.maxAsyncS3 || 20,
-      s3RetryCount: options.s3.s3RetryCount || 3,
-      s3RetryDelay: options.s3.s3RetryDelay || 1000,
-      multipartUploadThreshold: options.s3.multipartUploadThreshold ||  20971520,
-      multipartUploadSize: options.s3.multipartUploadSize || 15728640,
-      s3Options: {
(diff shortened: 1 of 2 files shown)


lib/plugins/s3/index.js
```diff
@@ -5,78 +5,75 @@ const s3 = require('s3');
...
+    const s3Options = options.s3;
+
+    throwIfMissing(s3Options, ['bucketname'], 's3');
     if (options.outputFolder) {
       throw new Error('The current version does\'t support S3 together with a configured outputFolder.');
     }
-    if (options.s3.key || options.s3.secret) {
-      throwIfMissing(options.s3, ['key', 'secret'], 's3');
+    if (s3Options.key || s3Options.secret) {
+      throwIfMissing(s3Options, ['key', 'secret'], 's3');
```
tobli: Is it really needed that both are passed as options? Or could one be passed as an option, and the other as an ENV variable? Perhaps not an interesting scenario in practise?

tobli:
@beenanner, @soulgalore - I'd appreciate your help testing/reviewing this. I wanted to separate out some cleanup before addressing the outputFolder "compatibility" for the S3 plugin.

soulgalore:
Have you tested it @tobli ? :)

tobli:
Guilty as charged! I have not, wasn't sure I could but partly out of laziness...

lib/plugins/s3/index.js
soulgalore: The use case is either you pass both or you have them in ENV, the combo seems not so important.

tobli: Hen maybe we can actually drop the check, just to have less code. Always setting the options should be fine, since they're often undefined.

soulgalore: we can take that one later on.

Issue #1539, opened 2017-03-21
Graphite >0.10.0 has different tag structure


kkopachev:
Graphite-web prior to 0.10.0 used to accept tags as strings. That's what currently used in [sitespeed.io](https://github.com/sitespeedio/sitespeed.io/blob/master/lib/plugins/graphite/send-annotation.js#L19)
However, [that's changed](http://graphite.readthedocs.io/en/latest/releases/0_10_0.html#id2), and that's going to 1.0.0 release (rc1 currently released). 

I tried few months ago with latest (at the moment) graphite-web, and it didn't accept tags as string.

soulgalore:
Hey @kkopachev thanks for informing us! If I understand the docs we could add a switch and then just make the tags an array instead of a string with commas. Just need to find a Docker container with 1.0.0-rc1 then so we can test.

Best
Peter

soulgalore:
I've pushed a branch that could fix it, could you test it @kkopachev ? Or do you know a Docker container that contains Graphite 1.0.0 rc1?

Best
Peter

soulgalore:
merged and will come in the next release.

PR #1540, opened 2017-03-21
Tags as array in Graphite 1.0 #1539

soulgalore:


lib/support/cli.js
```diff
@@ -211,6 +211,13 @@ module.exports.parseCommandLine = function parseCommandLine() {
       type: 'boolean',
       group: 'Graphite'
     })
+    .option('graphite.use1_0', {
```
soulgalore: we need a better name for this?

kkopachev:
That works, however in grafana shows up as one long tag. 
image

I didn't find good graphite docker image either, however I was able to combine few images together. I'll share with you soon so you can try too.

Graphite `/events`:
image

Grafana makes this query:
image

kkopachev:
I tested it a bit more and it looks like if you just pass tags as an array of strings, it would work as expected.
Here is my curl example:
```
curl -v -X POST http://localhost:8080/events/ -d \
'{"what": "Sitespeed.io", "tags" : ["cable","chrome","sitespeed_io","default","example_com","c_page"], "data": "some data"}'
```

soulgalore:
thank you @kkopachev for looking into it, I've did a fix, hopefully that will work.

lib/plugins/graphite/send-annotation.js
```diff
@@ -15,12 +15,14 @@ module.exports = {
     const connectivity = graphiteUtil.getConnectivity(options);
     const browser = options.browser;
     const namespace = options.graphite.namespace.split('.').join(',');
```
kkopachev: Wonder if that should be also individual elements of array?

soulgalore:
yes you are right let me fix that tomorrow.

soulgalore:
pre1 of Graphite is tagged so we should move on with this one. @tobli @beenanner any ideas about the naming?

kkopachev:
@soulgalore I didn't find good image with graphite+everything else in one. We're using a set of docker images with standalone services. I created similar setup https://github.com/kkopachev/sitespeed.io/tree/different-hosts 
One note here is that sitespeed assumes graphite-web and carbon run on the same host and only difference is port, which is not true in multi-container setup. So in order to test this PR I hacked `send-annotations.js` a bit.

lib/support/cli.js
tobli: Rather than specifying the release number, I'd name the option after the actual feature. Maybe graphite.arrayTags (defaulting to false to keep the current behaviour), alternatively graphite.stringTags (defaulting to true).

soulgalore:
@kkopachev cool, let us make it possible to configure different hosts.

PR #1542, opened 2017-03-22
Stop using options.urls[0] and start using the name of the run

soulgalore:
Then we can introduce a way of adding a name in next step?


tobli:
Breaking the dependency in plugins on options.urls is a great first step. 👍 👍 👍

PR #1553, opened 2017-03-28
Revert "Refactor result storage/url handling, and tweak S3"

soulgalore:
Reverts sitespeedio/sitespeed.io#1552


tobli:
We should be fine now with d5a301309af3837101fbbcad6a1694990a7498ea, including a bunch of unit tests.

PR #1563, opened 2017-04-04
First go going to tags in InfluxDB

soulgalore:
Is there another way we do this without rewriting the internal structure?
Do we have better names for tags? Thinking of "tools" "origin" and "summaryType"?


development/grafana-setup/entrypoint.sh
```diff
@@ -12,6 +13,9 @@ done
 
 echo "Up!"
 
+echo "Adding database for InfluxDB"
+curl -i -XPOST ${INFLUXDB_API} --data-urlencode "q=CREATE DATABASE sitespeedio"
```
tobli: I'm looking in [the documentation](https://hub.docker.com/_/influxdb/), and they use GET for creating a database, like this:
```
curl -G ${INFLUXDB_API} --data-urlencode "q=CREATE DATABASE sitespeedio"
```

soulgalore: The Docker README hasn't been updated to the latest InfluxDB version see https://docs.influxdata.com/influxdb/v1.2/guides/writing_data/#creating-a-database-using-the-http-api that is the correct way. GET is deprecated.

tobli: Cool, thankd

PR #1566, opened 2017-04-05
Use map and concurrency to make sure we don't overload the WPT server

soulgalore:
To fix these errors https://www.sitespeed.io/img/slack-alert-error.png

### DIFF
--- a/lib/plugins/webpagetest/analyzer.js
+++ b/lib/plugins/webpagetest/analyzer.js
@@ -30,7 +30,8 @@ module.exports = {
         log.verbose('Got JSON from WebPageTest :%:2j', data);
 
         const promises = [];
-        promises.push(wptClient.getHARDataAsync(id, {}));
+        let har;
+        promises.push(wptClient.getHARDataAsync(id, {}).then((theHar) => har = theHar));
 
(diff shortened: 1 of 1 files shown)


lib/plugins/webpagetest/analyzer.js
```diff
@@ -93,10 +94,12 @@ module.exports = {
 
           }
         })
-        return Promise.all(promises).then((result) => {
+
+        // set a low concurrency to not overload the WebPageTest server
+        return Promise.map(promises, function(){}, {concurrency: 2}).then(() => {
```
tobli: Sorry, but this will not have any effect. To quote the docs: 'The concurrency limit applies to Promises returned by the mapper function'. I.e. it waits on the promises returned by the mapper function, and keeps track of the number of currently unresolved promises. However this mapper function returns undefined, so the concurrency option will have no effect. Additionally since the async calls have already been made (when pushing the wptClient.get…Async result promises to the promises array), this doesn't affect the number of concurrent http calls.

To control this we need a queue, concurrent-queue as used for QueueHandler should work.

soulgalore:
darn!

PR #1567, opened 2017-04-05
Catch failing gets so we can generate the result page

soulgalore:


lib/plugins/webpagetest/analyzer.js
```diff
@@ -100,6 +100,14 @@ module.exports = {
           }
           myResult.trace = traces;
           return myResult;
+        }).catch(reason => {
```
tobli: I don't think this is the right approach. Promise.all will reject as soon as any promise is rejected, meaning it won't wait for the other ones. So we might return prematurely. Also, there's no knowing if the HAR capture was the one that failed or not.

I think a better approach is to append a ```.catch(WPTAPIError, (e) => …)``` to each promise we add to the promises array, if we want to indicate that a WPTAPIError is acceptable. The promise will resolve successfully, and we'd wait for them all in Promise.all. That way we also have the local context in the catch, i.e. we know what request failed. Building on that we could collect the errors and post an 'error' message to the queue. With this PR we'd only write the the log, but not record in the results (on the Errors page) that something went wrong.

soulgalore: Cool, lets do it like that!

Issue #1570, opened 2017-04-10
Optimize screenshot in results


pborreli:
My docker instance of grafana/sitespeed is running out of space because of the size taken by the screenshot in results page.

```bash
ls sitespeed-result/website.com/2017-04-10-10-04-06/pages/website.com/data/screenshots/

2.6M 0.png
3.0M 1.png
3.0M 2.png
```

Not sure is png is a good format for storing screenshots. 
What about optimising those files and maybe add an option to disable screenshots in results ?

lbod:
You can disable screenshots using `--plugins.disable screenshot`

pborreli:
@lbod thanks, indeed it doesn't generate the screenshot anymore but it generates 404 in results
![image](https://cloud.githubusercontent.com/assets/77759/24858690/455be74a-1de5-11e7-83b5-00eba7fdd9bf.png)

soulgalore:
In 3.x we used pngcrush if it was installed on the server to compress the image, lets do something like that again. Taking a PR for the broken image if you disable screenshots :)

soulgalore:
We should be able to do this with a plugin. We should make sure the plugin that stores the screenshots sends a message to the queue when the file is stored, and then this plugin will optimize the image.

pborreli:
shouldn't JPG a better format to store those captures ?

soulgalore:
yeah but then you need to have imagemagick or graphicmagick installed to convert it or is there another way?

Algram:
We could use something like Sharp https://github.com/lovell/sharp. It would still need compiling but has very good cross platform support and is fast. I like the idea of keeping this a plugin.

soulgalore:
@Algram yes good idea, that would be great to convert to jpeg + change the size. Let me give it a try.

soulgalore:
Did a quick try, this will work great. Think we should include it in the screenshot plugin by default and then make it configurable.

soulgalore:
Started in https://github.com/sitespeedio/sitespeed.io/tree/sharp

soulgalore:
This was released a couple of releases ago. You can now switch to jpg and set compression level.

PR #1571, opened 2017-04-10
Don't try to display unexistant screenshot

pborreli:
Removed broken image when screenshot plugin is disabled, aka `--plugins.disable screenshot`

Related to #1570


soulgalore:
thank you @pborreli !

PR #1574, opened 2017-04-16
Save metrics to disk when generating HTML

soulgalore:
If we store the raw JSON to disk that is used to generate the HTML, you can just run Pug command line and match the raw JSON with the pug template and debug/change the HTML on the fly, instead of having to run sitespeed.io over and over.

### DIFF
--- a/lib/plugins/html/htmlBuilder.js
+++ b/lib/plugins/html/htmlBuilder.js
@@ -22,6 +22,7 @@ class HTMLBuilder {
     this.storageManager = context.storageManager;
     this.timestamp = context.timestamp.format(TIME_FORMAT);
     this.options = options;
+    this.htmlOptions = options.html || {};
     this.dataCollection = context.dataCollection;
     this.summary = {};
     this.budget = context.budget;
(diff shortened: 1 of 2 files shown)


tobli:
We'll need to add all the helper functions when rerendering, right? I'm thinking a solution that replays incoming messages might be about the same
complexity to put together, and more full featured (reuse existing message save mechanism, can pick up new types of messages, allows for both changes to html and logic).

Could be this approach turns out really simple to do, but can we keep the PR open until the full setup is done? That way we only merge new options when we're sure it'll work.

soulgalore:
> We'll need to add all the helper functions when rerendering, right? 

No I thinking of another solution! Install pug-cli and then you can match the JSON file with the template and hack what you want really simple. Then you can do your changes without touching sitespeed.io.

soulgalore:
to make it clear: that way you can modify the pug on the fly and the html output will change when you change the pug.

tobli:
Sure, but calls to helpers (such as h.plural here https://github.com/sitespeedio/sitespeed.io/blob/master/lib/plugins/html/templates/runInfo.pug#L2) requires loading code from sitespeed.io, so just using pug-cli won't be enough.

Issue #1583, opened 2017-04-21
'outputFolder' option does not create the directory for the url's dns name.


abhagupta:
'outputFolder' option produces results but does not saves the result in a folder name of the url’s dns name. Earlier this feature was supported by `resultBaseDir` in version < 4 .   If I  understand correctly, `resultBaseDir` is not supported anymore, but how do get to store the results in same way in `outputFolder`?

soulgalore:
Yep I remember @tobli and I discussed but don't remember the outcome, I think we just missed it. Let us try to add it before 5.0 (we are still waiting in Selenium 3.4.0).

soulgalore:
discussed it with @tobli , we don't fully agree but there's a workaround that you can use, make sure you run sitespeed in the dir where you wanna put the data (when you run npm installed). For Docker it's just a matter of how you link in your volume, so you can choose where you wanna put your data. Else let us know if there's a use case for you where it doesn't work out!

abhagupta:
@soulgalore hmm  I understand. The problem is that I have created a wrapper which calls sitespeed to execute performance tests on android when user passes `android` in CLI to my wrapper, otherwise it just runs web performance tests.  Now, with web version, sitespeed runs in docker, so its easy to output the data in folder which I have exposed on a python simple webserver, so user can see their reports.  However, with android version (which is using non-docker sitespeed), I want to output the results in the same folder so that user can see their android results same way.  Otherwise i will have to start webserver for  2 locations.  I am open to suggestions on this approach as well.

PR #1584, opened 2017-04-22
Use resultBaseDir as in 3.x #1583

soulgalore:


lib/support/resultsStorage/index.js
```diff
@@ -26,7 +26,11 @@ module.exports = function(input, timestamp, outputFolder, resultBaseURL) {
     storageBasePath = path.resolve(outputFolder);
   } else {
     resultsSubFolders.push(getDomainOrFileName(input), timestamp.format('YYYY-MM-DD-HH-mm-ss'));
-    storageBasePath = path.resolve('sitespeed-result', ...resultsSubFolders);
+    if (resultBaseDir) {
+      storageBasePath = path.resolve(resultBaseDir,'sitespeed-result', ...resultsSubFolders);
```
tobli: According to the cli description, the sitespeed-result should not be added to this path.

lib/support/cli.js
```diff
@@ -521,7 +521,10 @@ module.exports.parseCommandLine = function parseCommandLine() {
       type: 'boolean'
     })
     .option('outputFolder', {
-      describe: 'The folder where the result will be stored.'
+      describe: 'The folder where the result will be stored, disabling hostname/date in the URL'
```
tobli: Maybe clearer as "hostname/date will not be appended to path"?

lib/support/resultsStorage/index.js
```diff
@@ -15,7 +15,7 @@ function getDomainOrFileName(input) {
   return domainOrFile;
 }
 
-module.exports = function(input, timestamp, outputFolder, resultBaseURL) {
+module.exports = function(input, timestamp, outputFolder, resultBaseURL, resultBaseDir) {
```
tobli: I'd prefer to keep the two path related parameters next to each other in the parameter list, and keep the URL last.

PR #1590, opened 2017-04-24
Update docs with docker details to remove data volumes

### LINKED ISSUE #1589: Clear docker graphite data
Hello,

Maybe this is not the best env to ask for this but couldn't find an answer anywhere.

I've setup my performance-dashboard via

```bash
curl -O https://raw.githubusercontent.com/sitespeedio/sitespeed.io/master/docker/docker-compose.yml
docker-compose up -d
```

And all works ok. But how do I clear the graphite data?

I tried to `docker-compose stop && docker-compose rm` and even deleted the entire images and restarted from scratch.

I still get the old data. How do I clear it?

radum:
If we close and remove the docker containers use for performance dashboard the data will be kept.

So on the next run it will persists with the new containers. To start from scratch we need to remove the volumes also.

This fixes #1589

### DIFF
--- a/development/index.md
+++ b/development/index.md
@@ -22,6 +22,12 @@ run the following to push the updated dashboards to Grafana.
 docker-compose up --build grafana-setup
 ```
 
+## Shutdown and remove containers
+
+When you are done you can shutdown and remove all the docker containers by running `docker-compose stop && docker-compose rm`
+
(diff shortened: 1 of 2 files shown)


soulgalore:
great thanks @radum !!!!

Issue #1087, opened 2016-07-27
Make it possible to link to specific tabs on page/run pages


soulgalore:
With the current solution you cannot choose which tab that is open when you access the URL the first time. We should find a way to fix that.

PR #1606, opened 2017-05-10
Added anchor fix scroll with the fixed navbar

radum:
The generated `toplist.html` page if it is long enough for the anchor scroll to work jumps to much because of the fixed navbar.

image

And this is a bit annoying, to fix it instead of using the negative margin / padding hack I preferred a more updated version using a small JS files that also takes into consideration the history using `history.pushState`

Adding this fix has this result:

image

This will fix the jump and the user will be able to move back and fwd with the browser buttons without a problem.

Also not sure if we should move the ID target for anchors to the actual heading instead of the table. I personally have no preference on this.

I only added the minified version of that file but here is the src for it:

```js
(function(document, history, location) {
  var HISTORY_SUPPORT = !!(history && history.pushState);

  var anchorScrolls = {
    ANCHOR_REGEX: /^#[^ ]+$/,
    OFFSET_HEIGHT_PX: 50,

    /**
     * Establish events, and fix initial scroll position if a hash is provided.
     */
    init: function() {
      this.scrollToCurrent();
      window.addEventListener('hashchange', this.scrollToCurrent.bind(this));
      document.body.addEventListener('click', this.delegateAnchors.bind(this));
    },

    /**
     * Return the offset amount to deduct from the normal scroll position.
     * Modify as appropriate to allow for dynamic calculations
     */
    getFixedOffset: function() {
      return this.OFFSET_HEIGHT_PX;
    },

    /**
     * If the provided href is an anchor which resolves to an element on the
     * page, scroll to it.
     * @param  {String} href
     * @return {Boolean} - Was the href an anchor.
     */
    scrollIfAnchor: function(href, pushToHistory) {
      var match, rect, anchorOffset;

      if(!this.ANCHOR_REGEX.test(href)) {
        return false;
      }

      match = document.getElementById(href.slice(1));

      if(match) {
        rect = match.getBoundingClientRect();
        anchorOffset = window.pageYOffset + rect.top - this.getFixedOffset();
        window.scrollTo(window.pageXOffset, anchorOffset);

        // Add the state to history as-per normal anchor links
        if(HISTORY_SUPPORT && pushToHistory) {
          history.pushState({}, document.title, location.pathname + href);
        }
      }

      return !!match;
    },

    /**
     * Attempt to scroll to the current location's hash.
     */
    scrollToCurrent: function() {
      this.scrollIfAnchor(window.location.hash);
    },

    /**
     * If the click event's target was an anchor, fix the scroll position.
     */
    delegateAnchors: function(e) {
      var elem = e.target;

      if(
        elem.nodeName === 'A' &&
        this.scrollIfAnchor(elem.getAttribute('href'), true)
      ) {
        e.preventDefault();
      }
    }
  };

  window.addEventListener(
    'DOMContentLoaded', anchorScrolls.init.bind(anchorScrolls)
  );
})(window.document, window.history, window.location);
```


soulgalore:
Hi @radum thanks for the PR, I think it worked with CSS before https://github.com/sitespeedio/sitespeed.io/issues/1087 but I could be wrong. Will try it out before the next release so it will go out then.

Best
Peter

radum:
@soulgalore Well even if we move the id to the heading element instead of the table it will still be cut, and doesn't look pretty :D

As for a css fix, apart for the negative margin fix (if we do this, we need to change background colors also) I can't think of any right now.

Cool, well let me know how it goes. Thanks

soulgalore:
Had a quick check and the old offender is 
https://github.com/sitespeedio/sitespeed.io/blob/master/lib/plugins/html/src/sass/components/tabs.scss#L35-L37, if we change that and adds the id:s on the h3:s instead, it should work just fine (I think). But I think you got more CSS skills than me so please check if you have time :)

radum:
@soulgalore Hmm, how did I missed that :) Yes that is the one. I will close this one then and try to follow the uses cases for when that thing was introduced and see how it goes if we remove it.

And I will move the id's to the headings.

soulgalore:
great thanks :)

PR #1608, opened 2017-05-11
Updated connectivity docs for TC to better explain what it will actually happen

radum:
Not sure what happened on my machine but my `-c 3g` setting is not working anymore, everything is fast as it would be on my 250 Mb connection.

Reading the docs I can see that I can set `--browsertime.connectivity.engine` to `tc` to solve this problem. Although I though tc is default now. (I'm running the latest docker version btw).

Anywho, the docs are not clear how will the download / upload speed be affected by tc so this is why I'm updating this to make it more clear on what happens with download when tc runs.


soulgalore:
Thanks @radum .  Best thing is always to user Docker network for now, thanks for the PR, will see if I can make the rest more clearer too!

soulgalore:
Actually the download speed will not be affected if you use tc, that was what I was trying to explain in the docs :) So lets make that more clear.

beenanner:
@radum you are right as `tc` is no longer the default and has been switch to `external` as you mentioned you need to specify the engine option if wanting to use `tc`. We can merge these changes and add to them to improve the documentation. As @soulgalore suggested we highly recommend docker as the preferred run option.

Issue #1613, opened 2017-05-16
New plugin - top list per page summary for 3rd party audit / tracking


beenanner:
Gather the following from a composite listing of all requests from every run for a particular page. 

- A list of the top 10 heaviest requests in reverse order.
- A list of the top 10 slowest requests in reverse order. 
- A list of the top 10 least performant requests in reverse order. This could be determined by Content Download (ms) / Bytes Downloaded = milliseconds per byte downloaded. 

Possible start with a tabbed HTML report for the pge summary report like we do for webpagetest and then allow sending of metrics to a TSDB for long term analysis. The TSDB will be tricky since for graphite it will eat up disk space for sites that have changing 3rd party resources. i.e. https://www.cnet.com/ We could document best practice for managing that data separately from the core site data. e.x. store for a significantly short time frame.

soulgalore:
The lists are almost the same as we have on the "toplist page", maybe we could add those lists to the current assets plugin, if you just configure <code>--firstParty</code>? That would work out fine I think. Sending to Graphite/InfluxDB, let us work on that. 

If you agree @beenanner let us setup another issue for what is needed for plugins to work as we want (add HTML/send to TSDB etc).

beenanner:
That sounds good to me @soulgalore. :-)

beenanner:
https://github.com/sitespeedio/sitespeed.io/issues/1141 for reference to plugin improvements. :-)

soulgalore:
We should re-factor assets a little, I think there are room for improvements and make it a little cleaner than today, I can start with that later this week.

soulgalore:
@beenanner made a shot at the first two lists, check if it is the way you want.

beenanner:
@soulgalore Nice! I'll check it out and see. Thanks!

soulgalore:
I think the mosquitoes important ones work, let us create.a new issue if we need more.

PR #1614, opened 2017-05-16
Slowest and heaviest 3rd party assets #1613

soulgalore:
We should look at those keys and check that it really works for Graphite/InfluxDB, haven't done that yet.


beenanner:
The output looks good. This is what I was testing with. I think the size matches a single asset but the response sizes seemed off even when calculating the sum of all the runs.

`bin/sitespeed.js -b chrome --firstParty ".*(cbsistatic|cnet).*" https://www.cnet.com/`

beenanner:
I did a single run let me know if I'm missing something. :-)

amgdgt com waterfall

vs

amgdgt com slowest

soulgalore:
That's a bug then that I think exists in PageXray, let us look into that.

soulgalore:
PageXray looks ok when I did a quick check, I think the problem is in PerfCascade, check the timings field, the total doesn't match, maybe introduced on 2.0:
screen shot 2017-05-16 at 17 16 53

soulgalore:
Ah ok, the error is in PageXray: We add both SSL and connect time to the total, but SSL is included in the SSL time.

soulgalore:
It should be fixed by 16cfbe5

soulgalore:
The group total that we used needs to be fixed, let me have a go at this later this week.

beenanner:
@soulgalore I also pushed a commit to link each asset to the page run that it originated from instead of the summary page run. This should make it easy to jump to the waterfall for deeper analysis.

beenanner:
For the group total do you mean the domain.html page? I can take a look just let me know. :-)

soulgalore:
Ah sorry @beenanner  that was more for me to remember. I kind of broke the current structure: we group every metric by domain, so that if you test multiple sites, the data are grouped by each site. The total is stats for all sites, and we only send individual groups to TDSB, so with the current structure we will not send the lists to TDSB (and that maybe is fine, let us discuss it).

soulgalore:
Ah I've forgot about this one, I'll see if I can fix this tomorrow.

soulgalore:
I'll have a go this weekend see if I can make it releasable, it is a nice feature.

soulgalore:
Think we are ready to go. We collect per group now but don't send/use it. If we ever need it, lets add it then and keep it as simple as possible for now.

PR #1623, opened 2017-05-23
Added support for --s3.storageClass option for user to be able to spe…

shakey2k2:
Added support for --s3.storageClass option for user to be able to specify the default storage class of the objects getting put into s3. Valid options are STANDARD | REDUCED_REDUNDANCY | STANDARD_IA

Really simple change. I was looking for a way to default the storage to reduced redundancy in s3 to lower costs. Let me know your thoughts.

Thanks

### DIFF
--- a/lib/plugins/s3/index.js
+++ b/lib/plugins/s3/index.js
@@ -43,7 +43,12 @@ module.exports = {
       localDir: baseDir,
       s3Params: {
         Bucket: s3Options.bucketname,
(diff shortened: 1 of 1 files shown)


shakey2k2:
Really simple change. I was looking for a way to default the storage to reduced redundancy in s3 to lower costs. Let me know your thoughts. 

Thanks

soulgalore:
thanks @shakey2k2 !

Issue #30, opened 2012-09-11
Rule: The number of javascripts


soulgalore:
Two or less is good, up to 4 means you should look into it and more than 4 means you need to fix it

PR #1644, opened 2017-06-20
WIP: Second try on adding Elasticsearch plugin

ftore:

- [x] I'm making a big change or adding functionality so I've already opened an issue proposing the change to other contributors, so I got feedback on the idea before took the time to write precious code: [#1062](https://github.com/sitespeedio/sitespeed.io/issues/1062)
- [ ] Update the documentation https://github.com/sitespeedio/sitespeed.io/tree/master/docs

### Description
In this pull request I implemented a plugin, which sends statistics data to Elasticsearch. Most aspiration was take from this pull request: [#1239](https://github.com/sitespeedio/sitespeed.io/pull/1239)

Right now, I am sending data without any modification. In later versions, I will create visuals and dashboards, and modify data based on those needs if necessary. In addition, I will add default dashboard in Kibana.

### How to run

First of all we'll need to start Elasticsearch and Kibana

```
cd tools/elasticsearch
docker-compose up
```

Then we'll need to activate the elasticsearch plugin and execute sitespeed.
```
bin/sitespeed.js --plugins.load elasticsearch --elasticsearch.host=localhost https://www.sitespeed.io/
```


soulgalore:
Hey @ftore I really like that you prepared a docker-compose file! 

If it is ok for you @beenanner, let us keep the PR and work with it the coming week, trying out dashboards and check if we wanna change the structure of the data, and then take the decision if we should include or have it as a plugin. I will have time to work/test it the coming weekend (also aiming for doing the next release at that time, but I'll test out the PR first).

ftore:
@peter, cool! I look forward to hearing from you back. Let me know if there is any issue

soulgalore:
@ftore one thing I'm thinking is that we should  start by justing sending the standard thing (runs from Browsertime and maybe WebPageTest and excludes lists as we do in the influxdb plugin: https://github.com/sitespeedio/sitespeed.io/blob/master/lib/plugins/influxdb/index.js#L35-L37 

and than we can add them later. What do you think about that?

Then in Graphite we inject a couple of things in the keys (browser, connectivity etc):https://github.com/sitespeedio/sitespeed.io/blob/master/lib/plugins/graphite/data-generator.js#L8-L34

And the same as tags for Influx:
https://github.com/sitespeedio/sitespeed.io/blob/master/lib/plugins/influxdb/data-generator.js#L29-L32

Thinking we should include that in ElasticSearch too?

ftore:
@peter sure, I will add them too. Initially, when I debugged the messages, I thought that messages are just flattened from JSON to string. I will exclude the lists and inject necessary things. However, should I create a new JSON message for connectivity and browser or inject to current JSON message? I guess injecting to existing JSON will be better, as we can search documents in ElasticSearch by connectivity and browser? What do you think?

soulgalore:
yes I think injecting should be good and try to keep it as simple as possible, but I haven't tried yet so not sure :)

ftore:
@peter, yesterday I debugged messages and found out that some messages have url and some don't. Those that have, that have in this format: http://www.sitespeed.io. I saw in graphite, url format is www_sitespeed_io. Should I transform it to www_sitespeed_io format? Those messages that don't have url, where can I get url from? I saw that options have it as an urls array, but I am not sure whether I can use this array.

beenanner:
For the URLs I would leave them.Graphite needs to transform due to it's storage design.

ftore:
ok, thank you @beenanner

ftore:
Why some messages have url value and some don't? For example message.type='browsertime.summary' doesn't have url defined..

soulgalore:
When a message have a URL specified, it's data is for that URL. **browsertime.summary** is a message containing data collected for all the urls, like the median firstPaint for all tested URL.

lib/plugins/elasticsearch/index.js
```diff
@@ -0,0 +1,58 @@
...
+    if (isEmpty(message.data))
+      return;
+
+    // TODO: we should add url values to all messages
+    // always have browser and connectivity in Browsertime and related tools
+    if (message.type.match(/(^pagexray|^coach|^browsertime|^largestassets|^slowestassets|^aggregateassets|^domains)/)) {
+
+      // if we have a friendly name for your connectivity, use that!
+      let connectivity = elasticsearchUtil.getConnectivity(this.options);
+
```
soulgalore: We have hack today so that users can give a friendly name to a URL. You can use it to call a long login URL to "login" or whatever you want, check: https://github.com/sitespeedio/sitespeed.io/blob/master/lib/support/tsdbUtil.js#L19-L27

I think we should check if we got a matching name and then also hack it into the message.

ftore: @soulgalore I don't understand this comment well. getConnectivity is returning native. But that hack you are mentioning is reformatting url. Could you please, be more specific? Also, could you please, give me more information about matching name?

soulgalore: Sorry I meant to show that it's not only connectivity that we can hack (adding extra info to the JSON object). When you test a URL from file you can add an alias:
`https://www.example.com/this/is/a/very/long/url long`

In InfluxDB & Graphite we use that alias as a friendly name (displayed in dropdowns etc) so that the gui will not break. So we should include that alias (if it exists) in the JSON blob too.

ftore: @soulgalore thanks for clarification. Updated.

beenanner:
@ftore I'm seeing a few types that have conflicts in kibana between indices. It appears they are around lastModifiedStats and expireStats between pagexray.pagesummary(number) and pagexray.summary (string). It appears we may not be casting those correctly on the pagexray side causing elasticsearch to store them differently?

pagexray.pageSummary
```
"expireStats": {
              "min": -6863971384173,
              "p10": 0,
              "median": 0,
              "p90": 21254400,
              "p99": 31536000,
              "max": 946080000
            },
```

vs 

pagexray.summary
```
"expireStats": {
              "median": "31536000",
              "mean": "28382460",
              "min": "600",
              "p90": "31536000",
              "max": "31536000"
            },
```

I'll file a ticket to see if we can correct that.

soulgalore:
@beenanner good find! we should fix that but then I think we shouldn't send *.summary to elasticsearch we should start with *.pageSummary and maybe send assets that we get from each run in browsertime, so we have each and every run stored (PageXray timing metrics is only for the first run if I remember correctly but we should veriffy that).

ftore:
@beenanner, did you file a ticket? I can try to fix them..

ftore:
@beenanner never mind, I found it. Here it is: [#30](https://github.com/sitespeedio/pagexray/issues/30)

ftore:
@beenanner, @soulgalore I think it is because toFixed function in [statsHelpers.js ](https://github.com/sitespeedio/sitespeed.io/blob/master/lib/support/statsHelpers.js#L52-L65) is returning string. If you look browsertime.summary, coach.summary, domains.summary, and pagexray.summary messages all have strings instead of numbers:

```
"median": "0",
            "mean": "0",
            "min": "0",
            "p90": "1",
            "max": "1"
```

soulgalore:
@ftore thanks for looking, I'll see if I can find the time the coming days (I'm on vacation) or if @beenanner can pick it up!

beenanner:
I'll take a look. Enjoy your vacation @soulgalore!

soulgalore:
hepp we should move on with this one now, I can start next week.

soulgalore:
so I didn't do anything here, I think we should make sure this is releasable to 6.0 sometimes in October.

raulbrito:
Hi, any news on the status of this PR? Thanks

soulgalore:
@raulbrito no status at all. I feel that no one of the core contributors got the knowledge of Elasticsearch, but if you feel you could take over and help, that would be great :)

soulgalore:
This is really old. If someone wants a elastic search plugin I think we should start from scratch since soooo many things changed since it was implemented.

Issue #1653, opened 2017-06-28
Add screenshots to Slack


soulgalore:
If we configured a resultBaseURL we should be able to include the path to the screenshot for the tested page in the Slack message, that would make it easier to see things fast.

soulgalore:
this was released before 6.0 :)

PR #1658, opened 2017-07-02
Add the screenshot to each attachment #1653

soulgalore:
I think this will do the thing we need, but let us merge in when we get close to 6.0.


soulgalore:
To get this to work, we need to send the Slack after the screenshot has been uploaded to the server.

PR #1659, opened 2017-07-02
Add charts to the result page.

soulgalore:
I think the visual progress chart is ok, but we need to work more on the page xray stuff.  need your feedback @beenanner when you have time and then we will keep this PR until 6.0.


beenanner:
@soulgalore looking pretty good! What about using pie charts for things that don't have multiple data series. i.e. request per type?

pie-chart

soulgalore:
@beenanner yes good idea!

soulgalore:
@beenanner did the styling of each request work out for you? When I tested all fields went black.

beenanner:
@soulgalore I'm pretty sure it worked for me. I can take another look. Do we have a specific color scheme we want to standardize on? WPT colors?

soulgalore:
@beenanner we can use the one we have in Grafana, they are from WebPageTest (so we all use the same :)). I think you re-added them the last time.

soulgalore:
I tried again @beenanner and for me it looks like this (testing cnet):
screen shot 2017-08-20 at 15 34 52

beenanner:
Let' me take a look @soulgalore. I was able to reproduce as well.

beenanner:
@soulgalore looks like the classes I added were the issue. I've reverted that commit for now, so colors should be back.  I can work on adding the grafana colors this week.

soulgalore:
Great @beenanner looks much better now. I need to do some fixes for visual progress that can be cleaner, will do that the coming week, then we could be ready to merge,

PR #1687, opened 2017-08-02
Fix for when wpt returns no data and a 200 response - Temp Fixes #1685

### LINKED ISSUE #1685: WebPageTest missing data breaks a run
We sometimes gets error when running WebPageTest against NYT:

screen shot 2017-07-31 at 06 56 22

The problem is that WebPageTest [misses data from one run](https://www.webpagetest.org/result/170730_W6_6c0ba8a9efbf1193ccdde74d950ed0d5/).

We should check for that in the JSON we get back.

https://www.webpagetest.org/jsonResult.php?test=170730_W6_6c0ba8a9efbf1193ccdde74d950ed0d5

beenanner:
Temp fix for #1685

### DIFF
--- a/lib/plugins/webpagetest/aggregator.js
+++ b/lib/plugins/webpagetest/aggregator.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const forEach = require('lodash.foreach'),
+  log = require('intel').getLogger('sitespeedio.plugin.webpagetest'),
   statsHelpers = require('../../support/statsHelpers');
 
 const metrics = ['TTFB', 'render', 'fullyLoaded', 'SpeedIndex'];
@@ -69,6 +70,8 @@ module.exports = {
             }
           });
         });
+      } else {
+        log.error('First View: Test Data Missing');
       }
     });
   },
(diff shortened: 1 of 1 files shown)


beenanner:
Better diff to ignore the whitespace. ;-)

https://github.com/sitespeedio/sitespeed.io/pull/1687/files?w=1

beenanner:
Hmm analyzer may be a better place for this patch.

soulgalore:
I think this is ok so we pickup the runs that worked? Do we also need to do a fix the template or does it work?

beenanner:
Let me do a bit more testing before we merge. Hopefully I can verify everything today. :-)

soulgalore:
What about adding a error log entry when it happens so people know?

beenanner:
In terms of the HTML everything is still working just no data. The screenshot error still surfaces in the logs, but if you think it should @soulgalore I can add another log entry for the zero data if the if condition fails. Let me know and I'll can update it tonight.

screen shot 2017-08-02 at 1 11 10 pm
screen shot 2017-08-02 at 1 11 16 pm

soulgalore:
cool, whatabout adding a log entry missing data or whatever the gui says in wpt?

beenanner:
Yes let me add that tonight then I'll merge this change for the next release.

soulgalore:
great!

PR #1711, opened 2017-09-15
Catch bad data in stats and metrics data.

tobli:


tobli:
Thx. We should look out for these log messages on the dashboard server.

Issue #1726, opened 2017-09-30
Allow plugins to know which other plugins configured


soulgalore:
Checkout https://github.com/sitespeedio/sitespeed.io/issues/1724

soulgalore:
Now they can talk to each other, keep it like that for now.

Issue #1724, opened 2017-09-28
Allow plugins to depend on other plugins


Mr0grog:
Is there any way to specify that one plugin requires another to be loaded? For example, if I’ve written a plugin that processes messages from gpsi in order to do something special, I’d like to ideally ensure that gpsi gets loaded without having to tell a user to make sure they’ve set `--plugins.load gpsi my-plugin`.

Barring that, it would at least be nice to have a final list of loaded plugins so I could throw an error or log a message noting that gpsi (or whatever other plugin I depend on) isn’t loaded.

Similarly, it would be helpful to be able to supply additional options to the plugin I depend on. For example, I am working on a plugin that requires a special post-script be used for Browsertime. It would be nice to somehow explicitly declare that I depend on Browsertime and provide the path to the post-script to load from my plugin instead of telling users to make sure they set that Browsertime option appropriately on the CLI.

soulgalore:
Hi @Mr0grog thanks for sharing, sorry that it took a couple of days to get back.

I'm gonna put in some work for the plugin structure for 6.0 sometimes in October or early November, but my main focus will be to add support for HTML (so you, when you do a plugin, can add your result to the HTML pages), for me that is the feature I miss the most.

Plugin dependency will make things more complicated so I'm scared of that.  One way to fix the problem for users is to supply [an example config.json](https://www.sitespeed.io/documentation/sitespeed.io/configuration/#configuration-as-json) where you can put all the config with the plugins that need to run. Maybe we could make the list of configured plugins available in the context, else I would add a check in close for my plugin and then log/throw if I didn't got any messages. Not perfect but it could be away to show it for the user.

> supply additional options to the plugin I depend on

Yep, we started this a year ago or something, like your plugin can say what parameters that they support so when you run --help the user sees those extra params? We should look at that again. I agree we should have something like that.

Best
Peter

Mr0grog:
> Plugin dependency will make things more complicated

Yeah, I can’t disagree with that! Even if explicit dependencies aren’t possible, it would be helpful if the the plugin `open` method received a list of all the loaded plugins so I could log a meaningful error telling the user exactly what they need to set for their `plugins.load` argument if the plugin I depend on isn’t present.

> like your plugin can say what parameters that they support so when you run --help the user sees those extra params?

[...]

PR #1728, opened 2017-09-30
Pass configured plugins as pluginNames in open #1726 #1724

soulgalore:
Let pass the array with configured plugin names so plugins can check which ones are configured.


soulgalore:
I think we can merge this before we branch of and start to work on 6.0?

PR #1730, opened 2017-10-02
Collect run/pageSummary from all plugins

soulgalore:
The current version has hardcoded plugins that reports run/pageSummaries but it's better to keep it open, it will make it much easier for plugins to get metrics into the HTML.

Let this be one of the first steps in 6.0.


soulgalore:
Lets rewrite data collection first

PR #1732, opened 2017-10-04
Switch from hooks to messages

soulgalore:
This is the first step to get rid of hooks:
* Introduce three new messages: setup, render and final

The setup message will make it possible for plugins to pass
configuration to each other (send pugs to the HTML plugin,
send javascript (that collects metrics) to Browsertime and can
hopefully, remove the post/preOpen hooks. The final message
replaces the postClose and final hook.


soulgalore:
We also need to switch to not use close() for the HTML plugin, we probably need more message types.

lib/support/queueHandler.js
```diff
@@ -159,13 +159,25 @@ class QueueHandler {
...
+   summarize - all analyze is finished and we can summarize all data
+   render - plugin store data to disk
+   final - is there anything you wanna do before sitespeed exist? upload files to the S3?
+   */
+    return this.startProcessingQueues()
+      .then(() => this.postMessage(make('setup')))
+      .then(() => this.drainAllQueues())
+      .then(() => Promise.map(sources, source => source.findUrls(this)))
       .then(() => this.drainAllQueues())
       .then(() => this.postMessage(make('summarize')))
```
soulgalore: Should we rename this? analyseFinshed?

soulgalore:
Lets go through what it looks like today (and what we wanna clean up)

## Close()
* The **HTML** plugin generates the HTML
* The **budget** plugin writes the result (JUnit/TAP)
* The **metric** plugin write down the metrics.txt file

## postOpen()
* The **metric** plugin setup the filters 

## postClose()
* The **s3** plugin sends the metrics to s3 (after the HTML is generated)

## final()
* The **slack** plugin slacks the result (after we have HTML and it exists om S3)

soulgalore:
I think the sane way would be to let each plugin collect the data it needs (so domain, assets, (new) pages) will collect everything and then in a one off shot send a message with all tha data, the HTML plugin will collect that and then automagically parse that data with the registered pugs. It need some changes in the naming of messages but it will make the overall structure cleaner.

soulgalore:
Renamed summarize to sitespeedio.summarize. Lets merge this and try out the new structure (IRL discussion with @tobli ):
1. Merge
2. Remove all actions on close
3. Create a core library to make it easier to see which support methods that needs to be exposed to plugins.
4. Rewrite the plugins to not require the support methods, instead get them in open.

tobli:
Here we go!

PR #1735, opened 2017-10-06
Graphite 1.x

soulgalore:
Moving to the official Graphite Docker container. Right now I've added the extra config files + the graphite.db so we map them = a little more work for the user but full flexibility we get rid off the burden of maintaining a Graphite docker container (+1m downloads).

What do you think @tobli @beenanner ?


soulgalore:
I've disabled the extra config files (you need them when you run in prod but not for testing it out). I'll add it to the docs. 

Should we make https://github.com/sitespeedio/sitespeed.io/blob/master/lib/support/cli.js#L293 default for 6?

tobli:
Looks good. Docker compose isn’t meant to be a production setup, so it’s fine to exclude some detailed config options. And graphite 1.0 style tags should be the default I think. As long as we mark it as a breaking change in the 6.0 release notes.

PR #1750, opened 2017-10-24
Added simple Jekyll search and document cleanup

soulgalore:
A lot of pages was missing descriptions, I've added some short texts but I think we have more 2do. 

@beenanner if you have time to try out the plugin this weekend it would be super, just hit the search in the header link. Feel free to push changed directly to the branch :)


soulgalore:
Lets merge this and we can test out the search, if we don't like it we can just disable it.

Issue #1681, opened 2017-07-26
Add output support for Google Lighthouse 


m0nkey:
SiteSpeed.io is already fantastic, and improving all the time. Out of the box support for Web Page Test and Google Page Speed Insights has been invaluable to compare and contrast different scorecards. Could support for output of Google Lighthouse CLI results be added to SiteSpeed.io reports?

m0nkey:
@tobli [suggested](https://twitter.com/tobiaslidskog/status/890280343722500097) using Browsertime to produce the trace logs, then running Lighthouse analysis on the results.

tobli:
Yes, I think that's the best way to start, since much of the work is already in place. Given the right options, the Browsertime plugin already posts trace messages for other plugins to digest. There might be some parts of Lighthouse we can't fully utilise, but I don't know enough to be sure.

For more reliable results, we only want to be running one browser instance at a time. That's enforced using the concurrency: 1 property on the Browsertime plugin. Decoupling the processing from that restriction will help keep things simpler. So if you'd, in theory, produce trace messages some other way (e.g. from files on disk) many traces could be analysed in parallel. That said it'd still be single threaded (that's how node works) but I'd be surprised if the processing becomes a bottleneck. The need for concurrency that's guiding the sitespeed.io architecture has primarily been for remote services and external processes (e.g. not blocking WPT calls when waiting for Chrome to launch locally).

m0nkey:
Cool, I didn't realize that regarding BT output traces (or haven't read the docs and code yet). As for what could or could not be compatible with Lighthouse stdin I don't really know yet either but I could do some investigation.

Re: concurrency it's probably OT for this thread but one example use case I was thinking of would involve running worker nodes to offload a queue or manifest of URLs. If I fed a list of 40 URLs to 4 workers, it would be could completed by those workers in parallel and 4x faster than serially.

siteriaitaliana:
I've tried the BT tracing approach and works with very few LH audits, main issue is:
https://github.com/sitespeedio/sitespeed.io/issues/1569
Probably add proper LH support in a plugin would be better to get the full advantages of it. If I make any progress with this I will create a PR.

soulgalore:
@siteriaitaliana has started on a plugin at https://github.com/siteriaitaliana/plugin-lighthouse help him out there :)

paveltsybenka:
@soulgalore  Peter, is there a chance to output metric values into log.
[...]

PR #1756, opened 2017-10-27
Added new LH plugin - plots_configs only

siteriaitaliana:

- [x] Update the documentation https://github.com/sitespeedio/sitespeed.io/tree/master/docs

### Description
This is an initial PR for #1681 . 
It contains a simple optional LH plugin (plots-config audits only for now  https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/config/plots-config.js). 

### Todo
- Expand the LH run to include at least all the Performance audits
- Store the json report in files
- Add few key score boxes in the html summary report
- Create a whole new LH html page

### DIFF
--- a/lib/plugins/lighthouse/aggregator.js
+++ b/lib/plugins/lighthouse/aggregator.js
@@ -1,0 +1,69 @@
+'use strict';
+
+const statsHelpers = require('../../support/statsHelpers');
+const METRIC_NAMES = [
+  'critical-request-chains',
+  'first-meaningful-paint',
+  'speed-index-metric',
+  'estimated-input-latency',
+  'first-interactive',
+  'consistently-interactive'
+];
+
+function replaceAll(str, search, replacement) {
+  return str.replace(new RegExp(search, 'g'), replacement);
+}
+
+module.exports = {
+  stats: {},
+  groups: {},
+  addToAggregate(data, group) {
+    if (this.groups[group] === undefined) {
+      this.groups[group] = {};
+    }
+    let stats = this.stats;
+    let groups = this.groups;
+
+    METRIC_NAMES.forEach(function(metric) {
+      if (metric == 'critical-request-chains') {
+        statsHelpers.pushGroupStats(
+          stats,
+          groups[group],
+          metric,
+          parseInt(data[metric].displayValue)
+        );
+      } else {
+        statsHelpers.pushGroupStats(
+          stats,
+          groups[group],
+          metric,
+          parseInt(data[metric].rawValue)
+        );
+      }
+    });
+  },
+  summarize() {
+    if (
+      Object.keys(this.stats).length === 0 ||
+      Object.keys(this.groups).length === 0
+    )
+      return undefined;
+
+    const summary = {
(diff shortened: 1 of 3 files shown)


soulgalore:
Hey Lorenzo, this is super cool and I think it will help many people but I want to keep it outside of sitespeed.io and have it as a standalone plugin. 

I will close the PR, the plan for 6.0 (I will do the most of the work in two weeks) is to add HTML capability to plugins so that we can build standalone plugins that can do the same as the core of sitespeed.io. When we release 6.0 we will also move out some of the current plugins to make it easier to maintain the core. I’ve just created https://github.com/sitespeedio/plugins  where we list all plugins.

When the foundation for 6.0 is ready, I will help you add the necessary stuff so it is ready when we release 6.0 (sometimes after FF 57 is released). Sorry that you spend time creating it but with 6.0 it will be a good standalone plugin and I will put that extra effort so everyone can use it asap when 6.0 is released.

There's one other reason why I don't want LH in sitespeed.io and this has annoyed me for a really long time. First: I think LH has some good measurements and can help you as a developer to run once in while on your site. But you need to know that the performance metrics in LH is not correct - the people at Google knows this but keeps pushing them and banging the drum before they really work. Let me explain some of them: 

SpeedIndex in LH is calculated using the frames/screenshots in devtools timeline. There's a [limit of 450 frames](https://bugs.chromium.org/p/chromium/issues/detail?id=739759) meaning if you have 60fps and a slowish site, everything will not be recorded. The only way to get correct metrics is to use WebPageTest or sitespeed.io/Browsertime.

FirstMeaningFul paint from Chrome is not correct for some sites at leasts, it happens earlier than first paint and that cannot be correct. See [this](https://github.com/WPO-Foundation/webpagetest/issues/942) and follow the links to Chrome bugs. I've been testing it out at work and I've seen it doesn't work yet. When it will, it will be super cool, but right now I don't no one should use it.

Different interactive metrics (TTI etc): I would wait until more browsers supported these metrics, how you calculate them etc. I understand the need for it, but I want to see more (or at least one) study about the metric.

Also to collect these kind of metrics you need to multiple runs (to get stable values) and have good way to set connectivity (WebPageTest use tc, sitespeed.io docker networks with tc etc), see https://github.com/GoogleChrome/lighthouse/blob/master/docs/throttling.md

I'll ping you ASAP when the structure for 6.0 is ready so we can add the HTML capability and your LH plugin will work out of the box.

Best
Peter

siteriaitaliana:
Agreed, still think few metrics in there are helpful though e.g. the critical request chain, and results look pretty stable as well with multiple cron runs
screen shot 2017-10-27 at 13 13 25
anyway, looking forward to the new 6.0 version! ;-)

soulgalore:
I think it's easy to be fooled by the scale on the left axis right? if you check in percentage for example speed index: (13650-9600)/9600 the diff compared with the lowest is 42%. Last night we switched from DO to AWS and what I'm seeing so far is 2,6% diff. But that's because AWS is so much more stable, so I wouldn't say that diff is dependent of the tool.

PR #1758, opened 2017-10-28
Remove hooks and use messages instead

soulgalore:
Let us iterate through this a couple of times (and test :)) before we merge. I think we should verify that this works by removing the hooks + in the next step clean up the code in each plugin.

### DIFF
--- a/lib/plugins/slack/index.js
+++ b/lib/plugins/slack/index.js
@@ -20,6 +20,70 @@ const defaultConfig = {
   limitMetric: 'coachScore'
 };
 
+function send(options) {
+  const slackOptions = merge({}, defaultConfig, options.slack);
+  const slack = new Slack(slackOptions.hookUrl);
+  const type = slackOptions.type;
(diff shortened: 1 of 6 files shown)


lib/plugins/slack/index.js
```diff
@@ -59,73 +60,76 @@ module.exports = {
...
-    const type = slackOptions.type;
-    const pageErrors = [];
-    const dataCollector = this.dataCollector;
-    let logo = 'https://www.sitespeed.io/img/slack/sitespeed-logo-slack.png';
 
-    let channel = slackOptions.channel;
-    if (channel && !channel.startsWith('#')) {
-      channel = `#${channel}`;
-    }
+      case 's3.finished': {
```
soulgalore: this need to be only s3 if s3 is configured, else we should send to html.finished.

soulgalore:
Verified and remove the hooks. We could do one more run of testing.

PR #1760, opened 2017-10-29
Feed plugins with messageMaker

soulgalore:
Feed plugins with messageMaker through context instead of require making it easier for plugin makers to send messages.


soulgalore:
Next step is to do the same with support/filterRegistry

PR #1764, opened 2017-10-29
Replace lodash.merge with Object.assign.

tobli:
Now that we require node 8, we don’t need lodash.merge anymore. Object.assign does the same thing.


tobli:
This was a bad idea. Object.assign doesn't recurse, and have recursive structures in a number of places. That may be a sign of complex design, but we'll leave it for now. Closing.

PR #1767, opened 2017-10-29
Datacollector break free

soulgalore:
Instead of having a generic dataCollector class: use the one we have and move it to the HTML plugin. Then refactor slack and text plugin to have lightweight versions.


tobli:
Free at last

This will be awesome!

PR #1779, opened 2017-11-06
If we get an error on the queue make sure we exit with an error #1267

soulgalore:


soulgalore:
Lets make a follow up PR that changes how we adds errors to the queue, so we name them. Would be nice to have different ones for BT and WPT etc.

soulgalore:
As discussed with @tobli : in the future move out the generic messages from the queue.

PR #1782, opened 2017-11-07
Upgrade to latest yargs

soulgalore:


soulgalore:
Let take this on later (and upgrade yargs to latest latest).

PR #1784, opened 2017-11-08
Let plugins register pugs = generate HTML

soulgalore:
There are three types (two implemented here):
* pug for the page summary page
* pug for a summary page

This commit moves out the pugs for WebPageTest and GPSI so we
have them as example plugins.


soulgalore:
Had a quick call with Tobias yesterday and we said it looked ok, lets merge it now and take on changes before we release 6.0 (if there are any).

PR #1786, opened 2017-11-09
First go at letting browsertime take scripts from other plugins

soulgalore:
Messages need to look like this:
```
queue.postMessage(
          make('browsertime.scripts', {
            category: 'superplugin',
            scripts: {
              userAgent: '(function() {return navigator.userAgent;})();',
              title: '(function() {return document.title;})();'
            }
          })
        );
```

And the will end up as a browserScript in the category. To have multiple categories you need to send multiple messages but that seems to be an ok limit.


beenanner:
Agreed multi messages for multiple categories seem like an acceptable design.

PR #1794, opened 2017-11-13
Plugin tab polish

tobli:
Tweaks to how plugins tabs are rendered in displayed. Fixes edge case crash.


soulgalore:
I think this PR changed the behavior? When I click on a tab, the browser scrolls down to the content, so clicking on tabs becomes a little jumpy (I tested in Chrome). Was that the intention? :)

soulgalore:
Looks better, great! One last thing: We miss the # in the URL now so you cannot share URL to specific tabs as before (see https://examples.sitespeed.io/5.0/2017-04-24-04-44-04/pages/en.wikipedia.org/wiki/Main_Page/index.html#browsertime). However maybe we should move to a "page per tab" solution, I'm not sure.

tobli:
Latest commit fixes the # fragment issue.

soulgalore:
perfect let us merge it tonight then.

PR #1799, opened 2017-11-18
Simplify storagemanager

tobli:
Remove possibility to pass promises for paths or data, this makes storage logic much easier.

Also push gzipping out to callers that actually use it. The code duplication is small, and imho worth it to keep the core code leaner.


soulgalore:
Puh this is great, the old structure made it hard to just store things :)

PR #1802, opened 2017-11-18
Put plugins tabs on correct row.

### LINKED ISSUE #1801: Tabs are broken for plugins since last tab fix
Checkout WebPageTest here:

screen shot 2017-11-18 at 22 28 28

https://results5.sitespeed.io/www.nytimes.com/2017-11-18-21-04-09/pages/www.nytimes.com/

It's the same for GPSI. Before they were added in a tab the same way as other plugins.

tobli:
A faulty indentation put plugin tabs outside the tabs div.

Fixes #1801


soulgalore:
WOW super fast fix :)

PR #1803, opened 2017-11-18
Move gpsi 

soulgalore:
Test out the new plugin at: https://github.com/sitespeedio/plugin-gpsi


tobli:
Awesome!
The html plugin still has references to gpsi in two places (start page boxes/detail page). Would be nice to get those out as well

soulgalore:
Yep fixed it now and added an issue for a generic solution in the future. One test case has a reference.

PR #1810, opened 2017-11-21
Register data for pug

soulgalore:
Now we use the id and type from the plugin, or should the plugin register the message type? @tobli


soulgalore:
@tobli let us sync: Either like this id +  type or send the specific full name, what do you think?

beenanner:
I think whichever would make it easiest on the plugin side would be best since we want to keep the barrier of entry as low as possible for developers?

soulgalore:
Yeah I agree but what is simplest? I'm not sure.

Issue #1778, opened 2017-11-04
FTP uploader plugin


moix:
Hi!
I wanted to contribute with a simple ftp uploader plugin to publish the output of the executions into an ftp, I find it useful, so I send a pr some minutes ago, #1777, however the build failed. I'd be happy to fix it if is related, anyone can help here?

Thanks!

soulgalore:
Let us create a new issue when this is needed. I think now when we both support s3 and the google version have what we need.

PR #1777, opened 2017-11-04
FTP uploader plugin

moix:
This simple plugin add the capability to upload output to a ftp

PR #1821, opened 2017-11-23
FTP uploader plugin (#1778)

moix:
This simple plugin add the capability to upload output to a ftp (#1778)


moix:
ups sorry for the mess, couldn't work on this for a while and trying to fix the conflicts I recreated the pr already requested on https://github.com/sitespeedio/sitespeed.io/pull/1777
checks should pass now I couldn't identify what is failing now in travis related to these changes. tx

soulgalore:
Cool thanks, I have look later, will merge it as soon as we released 6.0 (tomorrow maybe) and the we can ship it in 6.1.

soulgalore:
@beenanner @tobli when someone else submits a PR the WPT key fails, do you have any idea of why?

tobli:
The WPT failure is due to https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml. PRs from forks don't have access to encrypted environment variables. We should make the wpt check conditional, I think $TRAVIS_PULL_REQUEST_SLUG should work, if it's empty or "sitespeedio/sitespeed.io" we're from within the sitespeed.io repo.

soulgalore:
@tobli ah nice, lets see if we can fix that.

tobli:
Should be fixed with https://github.com/sitespeedio/sitespeed.io/commit/ced6252b01734f99c38989280027c3929e2545bb

soulgalore:
Seems to be something more to fix @tobli got an error from the oneliner

soulgalore:
Lets add a test in travis for the ftp, I think something like:
```
pip install twisted
twistd ftp --auth=memory:myuser:mypasswd
```
should work to setup a ftp locally?

lib/plugins/ftp/index.js
```diff
@@ -0,0 +1,83 @@
...
+      const baseDir = this.storageManager.getBaseDir();
+      const client = new PromiseFtp();
+
+      return new Promise(resolve => {
+        log.info(
+          `FTP uploading output (${baseDir}) to FTP ${ftpConfig.user}@${
+            ftpConfig.host
+          }, this can take a while ...`
+        );
+        var dirlist = [];
```
soulgalore: All var can be const if you don't need to re-assign them.

lib/plugins/ftp/index.js
```diff
@@ -0,0 +1,83 @@
...
+              filelist.push(path.join(dir, file));
+            }
+          });
+        }
+
+        walkSync(baseDir);
+        client
+          .connect(ftpConfig)
+          .then(function() {
+            dirlist.forEach(function(element) {
```
soulgalore: You need to wait for all promises here, look at http://bluebirdjs.com/docs/api/promise.all.html for example and then when all the dirs is created, you can upload the file.

lib/plugins/ftp/index.js
```diff
@@ -0,0 +1,83 @@
...
+              log.debug(`FTPying directory(${element})`);
+              return client.mkdir(element, true);
+            });
+            filelist.forEach(function(element) {
```
soulgalore: Same here: http://bluebirdjs.com/docs/api/promise.all.html

lib/plugins/ftp/index.js
```diff
@@ -0,0 +1,83 @@
...
+              return client.put(element, element);
+            });
+          })
+          .then(function() {
+            if (ftpConfig.removeLocalResult) {
+              fs.remove(baseDir).then(() => {
+                log.debug(`Removed local files and directory ${baseDir}`);
+              });
+            }
+            queue.postMessage(make('ftp.finished'));
```
soulgalore: I think this should wait on remove, like right now if you remove the dirs, ftp.ifnished will be fired before the files are removed.

lib/plugins/ftp/index.js
```diff
@@ -0,0 +1,83 @@
...
+            return client.end();
+          })
+          .then(function() {
```
soulgalore: Just FYI you can rewrite these to => to have less code.

soulgalore:
Closing this since there where no changes.

PR #1832, opened 2017-11-27
Add/remove is converted in https://github.com/sitespeedio/sitespeed.io/blob/master/lib/sitespeed.js#L94-L95

### LINKED ISSUE #1831: Repeating `--plugins.add` or `--plugins.remove` throws exceptions
I’ve just upgraded to v6 and the configuration documentation says “specify multiple plugin names separated by comma, or repeat the --plugins.add option” for both `--plugins.add` and `--plugins.remove`. However, repeating them doesn’t actually work and in fact throws an exception:

```sh
$ bin/sitespeed.js http://www.sitespeed.io --plugins.add plugin-1 --plugins.add plugin-2
> plugins.add.split is not a function
```

It looks like the problem is that the [`coerce` function for plugins doesn’t check the type for the `add`/`remove` properties](https://github.com/sitespeedio/sitespeed.io/blob/master/lib/cli/cli.js#L641-L649). It should probably also make sure to split each item in the array (if the properties are arrays) and flatten them, since it could be repeated with comma-delimited lists, too:

```sh
$ bin/sitespeed.js http://www.sitespeed.io --plugins.add plugin-1 --plugins.add plugin-2,plugin-3
```

soulgalore:
Fixes #1831 

@tobli do you remember, did we need the coerce in cli for anything, maybe I'm missing something.


Mr0grog:
I think you still need it because the value of `plugins.add` and `plugins.remove` after parsing could be either a string or array, e.g:

```sh
# This will result in a string:
bin/sitespeed.js http://www.sitespeed.io --plugins.add plugin-1,plugin-2

# But this will result in an array:
bin/sitespeed.js http://www.sitespeed.io --plugins.add plugin-1 --plugins.add plugin-2
```

soulgalore:
yes but that is fixed later on in the code https://github.com/sitespeedio/sitespeed.io/blob/master/lib/sitespeed.js#L94-L95 but lets see.

Mr0grog:
This definitely breaks the comma-separated case:

```sh
$ bin/sitespeed.js http://www.sitespeed.io -n 1 --plugins.remove coach,budget --plugins.list

[2017-11-27 14:49:19] INFO: Versions OS: darwin 16.7.0 nodejs: v8.9.1 sitespeed.io: 6.0.1 browsertime: 2.0.0 coach: 1.1.0
[2017-11-27 14:49:19] INFO: The following plugins are enabled: assets, browsertime, budget, coach, domains, harstorer, html, metrics, screenshot, text, tracestorer
[2017-11-27 14:49:19] INFO: Starting chrome for analysing http://www.sitespeed.io 1 time(s)
[2017-11-27 14:49:19] INFO: Testing url http://www.sitespeed.io run 1
[2017-11-27 14:49:26] INFO: 11 requests, 177.37 kb, backEndTime: 201ms, firstPaint: 434ms, DOMContentLoaded: 217ms, Load: 431ms, rumSpeedIndex: 434
[2017-11-27 14:49:28] INFO: HTML stored in /Users/rbrackett/Dev/sitespeed.io/sitespeed-result/www.sitespeed.io/2017-11-27-14-49-19
[2017-11-27 14:49:28] INFO: Finished analysing http://www.sitespeed.io
```

Note that `coach` and `budget` were both removed, but they’re still listed here.

soulgalore:
yes you are right, I didn't know we had a comma separated case :|

Mr0grog:
Sorry, I should have been more clear in my original response—was trying to point that out.

PR #1838, opened 2017-11-28
Use sharp to change screenshot size or convert to JPG [Merge for 6.1]

soulgalore:
We need to update the Dockerfile to include build-essential (higher up in the Docker chain).


lib/plugins/screenshot/index.js
```diff
@@ -1,46 +1,98 @@
...
-      screenshot.name,
-      url,
-      'screenshots'
-    )
-  );
+function storeScreenshots(url, imagesAndName, storageManager, options) {
+  return Promise.map(imagesAndName, screenshot => {
+    const maxSize = get(options, 'maxSize', defaultConfig.maxSize);
+    const imageType = get(options, 'type', defaultConfig.type);
+    if (imageType === 'png') {
```
tobli: Do this if once, outside the map. Then just set an outputOptions object, to be passed to sharp.

lib/plugins/screenshot/index.js
```diff
@@ -1,46 +1,98 @@
...
+      const compressionLevel = get(
+        options,
+        'png.compressionLevel',
+        defaultConfig.png.compressionLevel
+      );
+      sharp(screenshot.data)
+        .png({ compressionLevel })
```
tobli: Use [toFormat(imageType, outputOptions)](http://sharp.dimens.io/en/stable/api-output/#toformat) instead to remove the if/else.

lib/plugins/screenshot/index.js
```diff
@@ -1,46 +1,98 @@
...
+        .resize(maxSize, maxSize)
+        .max()
+        .toBuffer()
+        .then(newBuff => {
```
tobli: this could be simplified to skip the { and return.

lib/plugins/browsertime/index.js
```diff
@@ -271,10 +271,17 @@ module.exports = {
           .tap(results => {
             if (results.screenshots) {
               queue.postMessage(
-                make('browsertime.screenshot', results.screenshots, {
-                  url,
-                  group
-                })
+                make(
+                  'browsertime.screenshot',
+                  {
```
tobli: This is technically a backwards incompatible change (external plugins would need to change as we do). I think it's ok, just need to include a mention in the changelog (in the PR would be good).

A good reminder as well to audit all other plugins to always use a top level object for message data, to allow for further expansion without similar api breakage.

lib/plugins/html/templates/url/index.pug
```diff
@@ -96,8 +96,9 @@ block content
         .one-half.column
             if hasScreenShots
               - var width = options.mobile ? 150 : '100%';
-              a(href='data/screenshots/0.png')
-                img.screenshot(src='data/screenshots/0.png', width=width, alt='Screenshot of run 1')
+              - var screenshotName = 'data/screenshots/0.' + screenShotType
+              a(href=screenshotName)
+                img.screenshot(src=screenshotName , width=width, alt='Screenshot of run 1')
```
tobli: remove the space after screenshotName here.

lib/plugins/browsertime/index.js
soulgalore: hmm that's true we should do it in another way.

lib/plugins/screenshot/index.js
```diff
@@ -1,46 +1,98 @@
...
+  return Promise.map(imagesAndName, screenshot => {
+    const maxSize = get(options, 'maxSize', defaultConfig.maxSize);
+    const imageType = get(options, 'type', defaultConfig.type);
+    if (imageType === 'png') {
+      const compressionLevel = get(
+        options,
+        'png.compressionLevel',
+        defaultConfig.png.compressionLevel
+      );
+      sharp(screenshot.data)
```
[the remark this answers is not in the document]
soulgalore: fixed

lib/plugins/screenshot/index.js
```diff
@@ -1,46 +1,98 @@
...
+          return storageManager.writeDataForUrl(
+            newBuff,
+            screenshot.name,
+            url,
+            'screenshots'
+          );
+        });
+    } else {
+      const quality = get(options, 'jpg.quality', defaultConfig.jpg.quality);
+      sharp(screenshot.data)
```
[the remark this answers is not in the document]
soulgalore: fixed

lib/plugins/screenshot/index.js
```diff
@@ -1,46 +1,98 @@
...
-      screenshot.data,
-      screenshot.name,
-      url,
-      'screenshots'
-    )
-  );
+function storeScreenshots(url, imagesAndName, storageManager, options) {
+  return Promise.map(imagesAndName, screenshot => {
+    const maxSize = get(options, 'maxSize', defaultConfig.maxSize);
+    const imageType = get(options, 'type', defaultConfig.type);
```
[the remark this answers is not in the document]
soulgalore: fixed

lib/plugins/screenshot/index.js
soulgalore: fixed

lib/plugins/screenshot/index.js
soulgalore: fixed

lib/plugins/screenshot/index.js
```diff
@@ -1,46 +1,98 @@
...
+    }
+  });
 }
 
 module.exports = {
   open(context, options) {
     this.storageManager = context.storageManager;
     this.make = context.messageMaker('screenshot').make;
-    this.options = options;
+    this.options = options.screenshot;
```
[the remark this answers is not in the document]
soulgalore: fixed (added specific screenshot config).

lib/plugins/screenshot/index.js
```diff
@@ -1,46 +1,98 @@
...
+    this.imageType = get(this.options, 'type', defaultConfig.type);
```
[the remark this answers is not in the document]
soulgalore: fixed.

tobli:
👍 Looks good to go with the changes (I didn't actually test it myself). Not sure if the message data compatibility is a blocker.

soulgalore:
We could change so it is the actual screenshot plugin that adds the types, hmm.

lib/plugins/browsertime/index.js
soulgalore: I've changed it now so that HTML listen on the browsertime config.

soulgalore:
This should be good to go now.

Issue #1474, opened 2017-02-26
Compress files before we upload them to S3


soulgalore:
It would be nice to save some money and make it possible to compress files before we upload them, and then make sure S3 serves them https://zanon.io/posts/serving-gzipped-files-in-amazon-s3-cloudfront

soulgalore:
This is doable now the the HTML plugin sends a message when it is finished **html.finished**

soulgalore:
We now gzip most files, think that is enough.

PR #1839, opened 2017-11-28
Compress HTML files #1474 [Merge for 6.1]

soulgalore:
Made a first go at compressing HTML for be served from S3 using something like https://zanon.io/posts/serving-gzipped-files-in-amazon-s3-cloudfront

We also need to compress CSS/JavaScript. Is that something one of you wanna have a go at?

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -561,6 +561,11 @@ module.exports.parseCommandLine = function parseCommandLine() {
       default: 10,
       group: 'HTML'
     })
+    .option('html.gzip', {
+      describe: 'Gzip all HTML/assets files. Handy when serving from S3.',
+      default: false,
+      group: 'HTML'
+    })
     .option('summary', {
       describe: 'Show brief text summary to stdout',
       default: false,
--- a/lib/plugins/html/htmlBuilder.js
+++ b/lib/plugins/html/htmlBuilder.js
@@ -10,7 +10,8 @@ const helpers = require('../../support/helpers'),
   packageInfo = require('../../../package'),
   renderer = require('./renderer'),
   metricHelper = require('./metricHelper'),
-  isEmpty = require('lodash.isempty');
+  isEmpty = require('lodash.isempty'),
+  zlib = require('zlib');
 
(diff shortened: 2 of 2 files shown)


tobli:
Are we sure we want this? When the article was written CloudFront (the AWS CDN used in front of S3) didn't support automatic gzipping, but it was added in December 2015.

soulgalore:
I think the win is not for sending gzip to the browser (that post is showing how we need to do it), gzipping for saving money on S3 is the thing I think.

tobli:
I don't think that saving is worth it. S3 is $0.023/GB/month, so you'd need to save 43 GB by gzipping before upload in order to save $1/month. Compressing adds complexity, and makes the html site not browsable locally.

soulgalore:
I don't agree, I'm from the working class, every dollar counts :) 

We pay something like $15 per month last time I checked with dashboard.sitespeed.io. 188Gb and then we only store the data for 60 days. And we are not a heavy user, so I think the overall save for sitespeed.io users can be big. 

But I'm open to suggestions, maybe there's a better way we can fix this.

kkopachev:
I also had such idea of gzipping files right before uploading to s3, but realistically, heaviest assets on each run is probably video recording and screenshots. How much real savings is here? 
Not sure what exactly lives on dashboard.sitespeed.io, but it seams that results stored on S3 under  results.sitespeed.io.

On implementation prospective, why do you want to store `.gz` everything? From my understanding it would be easier to compress text files right before uploading to S3 and using normal extensions (like `.html`) and setting appropriate headers. 

I also noticed, that har file served gzipped, and then js library decodes content. Can't browser do that by itself, if given `Content-Encoding` header?

soulgalore:
Agree that it makes things more complicated. Another way could be to do it as a standalone plugin that either gzip recursively or zip the whole folder (and open for WebPageTest like S3 storing).

>  How much real savings is here?
It depends I think. Now when you can configure the video to be smallish  (low quality) and change the images with sharp (when we merge this PR) it can be a larger part, but I agree it would be nice to have some numbers. Lets wait with this and see if we have another approach.

PR #1840, opened 2017-11-28
Don’t test WPT on Travis for PRs from forks.

tobli:
Forks don’t have access to encrypted environment variables on Travis, so referensing $WPTKEY doesn’t work. Instead just run these tests in the sitespeedio/sitespeed.io repo.


soulgalore:
Hmm this is strange, it worked in the PR but makes other PRs fail, reverting for now.

PR #1871, opened 2018-01-09
Add wpt metrics

jpvincent:

- [x] Update the documentation https://github.com/sitespeedio/sitespeed.io/tree/master/docs

### Description

Following the new WPT grafana dashboards (https://github.com/sitespeedio/grafana-bootstrap-docker/pull/14), and as requested by @soulgalore here is my proposal to add more metrics from WPT to grafana.

For the demo to work correctly, please activate the `webpagetest.timeline` option (was already there) and the newly added `webpagetest.domainsDashboard`.
I also updated the documentation accordingly, tell me if all looks right.


docs/documentation/sitespeed.io/configuration/config.md
```diff
@@ -94,6 +94,9 @@ WebPageTest
   --webpagetest.script             The WebPageTest script as a string.
   --webpagetest.includeRepeatView  Do repeat or single views  [boolean] [default: false]
   --webpagetest.private            Wanna keep the runs private or not  [boolean] [default: true]
+  --webpagetest.timeline           Activates chrome tracing. More Useful metrics but can slow down the test itself  [boolean] [default: false]
+  --webpagetest.domainsDashboard   Record weight and requests for each domain to populate the grafana dashboard. Will take more disk space in graphite [boolean] [default: false]
```
soulgalore: We send over all webpagetest.* metrics to the WebPageTest API, I think maybe we should mark the ones that are sitespeed.io own, so it makes it easier to understand. @tobli @beenanner any input?

soulgalore:
@jpvincent Looks great, thanks this will be really cool! Only thing is that extra parameter, let us sync that and then merge.

docs/documentation/sitespeed.io/configuration/config.md
soulgalore: I mean *webpagetest.domainsDashboard* :)

jpvincent: Yeah I thought about deleting the option once it has been used in `index.js`, but modifying the options directly is never a good idea.

Later on, I would like to add a connection alias (rather than "custom") for WPT, so the same problem will occur.

Maybe we should add a sub namespace, like
```
--webpagetest.ssio.enableDomains=…
--webpagetest.ssio.connectivityAlias=…
```

soulgalore: Yes, either that or adding *--webpagetestParams* (following the same patterns as we have for some others)? The thing with the WebPageTest NodeJS API is that it changes options passed into it but if I remember correctly we "fixed" that by passing a copy of the object to the api.

Thinking some more, if you add the .ssio and then we could remove it from the cloned version here: https://github.com/sitespeedio/sitespeed.io/blob/master/lib/plugins/webpagetest/analyzer.js#L23

jpvincent: you're the architect, tell me what would fit best

soulgalore: I want input from @beenanner and @tobli :)

soulgalore:
Let me merge it, add the sub name space and then let it run for a while on dashboard.sitespeed.io and we can add your new dashboards.

soulgalore:
Thanks again @jpvincent of for the cool contribution :)

soulgalore:
I've put the domainsDashboard under -webpagetest.ssio.domainsDashboard but keep it out of the docs and cli for now (like a hidden feature).

lib/plugins/webpagetest/index.js
```diff
@@ -13,18 +13,39 @@ const fs = require('fs');
...
+  'data.median.firstView.score_combine',
+  'data.median.firstView.score_minify',
+  'data.median.firstView.domElements',
+  'data.median.firstView.lastVisualChange',
+  'data.median.firstView.visualComplete85',
+  'data.median.firstView.visualComplete90',
+  'data.median.firstView.visualComplete95',
+  'data.median.firstView.visualComplete99',
+  'data.median.firstView.FirstInteractive',
+  'data.median.firstView.TimeToInteractive',
```
wolframkriesing: does this prevent the data from repeat view to be processed? because it explicitly says `data.media.firstView...` now, instead of `data.median.*...`

soulgalore: ooops, you are right, we changed the default behavior there, let me fix that. Usually I only test first view but we shouldn't change how we collect metrics.

PR #1884, opened 2018-01-15
Send annotations if you run WebPageTest standalone.

soulgalore:
Before we only sent annotations for Browsertime but if you run WebPageTest standalone we should also send annotations.

I haven't actually tested this yet so let us take it easy with merging until we verified all cases work.

Ping @jpvincent


soulgalore:
@tobli @beenanner need your eyes on the if-statement so we are sure it works out :)

soulgalore:
Yes great, I've pushed the changes.

PR #1903, opened 2018-02-04
Improve introduction readibility

DebugXYZ:
Stumbled a bit when reading the introduction documentation, think these updates might help.

I think prefixing "It should" instead of "It" makes it easier to read without rephrasing too much. Also updated the last bullet point on what I believe to be the intent.


soulgalore:
Thank you @DebugXYZ much better!

PR #1679, opened 2017-07-24
add extra S3 settings

adrien-f:
### Your checklist for a pull request to sitespeed.io
Please review the [guidelines for contributing](CONTRIBUTING.md) to this repository.

- [ ] I'm making a big change or adding functionality so I've already opened an issue proposing the change to other contributors, so I got feedback on the idea before took the time to write precious code
- [X] Check that your change/fix has corresponding unit tests (if applicable)
- [X] Squash commits so it looks sane
- [X] Update the documentation https://github.com/sitespeedio/sitespeed.io/tree/master/docs
- [X] Verify that the test works by running `npm test` and test linting by `npm run lint`

### Description

Greetings,

This is a small pull request to add more settings to the S3 Upload plugin, I needed to have the possibility to update the AWS endpoint (I'm using Ceph instead), disable the SSL and set the ACL when uploading.

Let me know if you need more documentation updated or if there's a question.

PR #1937, opened 2018-03-08
Add option to S3 plugin for ACL settings on put objects, this setting…

### LINKED ISSUE #1935: Add S3 ACL settings as an option
I have a fix for this and can submit a PR (we are meant to submit an issue first?). 

No canned ACL is set on the S3 objects created and currently by default we have to manually change them in the console if we need to, by default I'm certain they are set to private (at least in our setup, they don't inherit bucket policy). It's a trivial task to add the option i.e. https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl and I've tested this locally and it worked. 
I've been defensive with the `s3Params` object rather than leave it undefined or an empty string or guess what the default is in S3. I tried to use a spread operator to optionally add the property in the literal but linting didn't allow, it's not important but just wanted to explain the simple code. 

I can submit the pr now if ok

lbod:
This option currently isn't provided by the s3 plugin.
The option `s3.acl` allows adding one of the canned ACL's for S3 buckets rather than letting a default setting. See #1935  

Closes #1935 

### Description

### DIFF
--- a/docs/documentation/sitespeed.io/configuration/config.md
+++ b/docs/documentation/sitespeed.io/configuration/config.md
@@ -111,6 +111,7 @@ s3
   --s3.bucketname         Name of the S3 bucket
   --s3.path               Override the default folder path in the bucket where the results are uploaded. By default it's "DOMAIN_OR_FILENAME/TIMESTAMP", or the name of the folder if --outputFolder is specified.
   --s3.region             The S3 region. Optional depending on your settings.
+  --s3.acl                The S3 canned ACL to set, optional depending on your settings.
   --s3.removeLocalResult  Remove all the local result files after they have been uploaded to S3  [boolean] [default: false]
 
 HTML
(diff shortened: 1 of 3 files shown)


soulgalore:
Thanks @lbod I'll merge first thing tomorrow. Also I wanna make https://github.com/sitespeedio/sitespeed.io/pull/1679#issuecomment-324304106  happened so we have a generic way of adding extra fields.

lbod:
That's cool, dynamic settings would be much better!

That travis build failed but it doesn't look like it's anything to do with my commit (webpage test api key) https://travis-ci.org/sitespeedio/sitespeed.io/builds/351052423#L2402

soulgalore:
@lbod no we should exclude the WPT run for PRs, @tobli did a quick hack but we couldn't get it to work.

Thanks for adding the extra info etc for the PR. I'll merge but the I'll see when I can get around to test the generic AWS config.

soulgalore:
@lbod fyi the config docs is automated generated when we do a release. Sorry that is probably not documented. It works like this:
- We always push master of the docs to https://www.sitespeed.io
- When we release browsertime/sitespeed.io the configuration file is autogenerated from latest version, and we push it. We have it that way so that we only update the docs when we get a release. That's why we also separate documentation changes for new things in another PR that we merge just before we do the release.

Thanks for the PR!

PR #1946, opened 2018-03-14
separate tests in Travis

soulgalore:


travis/wptTests.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -ev
+if [ "${WPTKEY}" != "false" ]; then
```
tobli: What scenario is this meant to cover? Is it perhaps "$TRAVIS_PULL_REQUEST" != "false", as described by
https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions ?

soulgalore:
No I think the security variables (like the wpt config) will not be availible for users that atre not in the team when they do a PR? The generic PR check will make the teams PR also miss running WPT?  Just looking for a way to make users PRs pass and teams PR run all the test.

Issue #1959, opened 2018-03-24
Docker ffprobe Path Using Full URI


rsiddle:
The following URL works on my Mac (6.4.0), but seems to fail on Docker (6.5.0). I've tried variations around Docker on 6.4.X too, with the same issue.

Example URL: http://www.argos.co.uk/browse/technology/ipod-mp3-and-headphones/c:29957/

```
docker run --shm-size=1g --rm -v "$(pwd)":/sitespeed.io sitespeedio/sitespeed.io:6.5.0 'http://www.argos.co.uk/browse/technology/ipod-mp3-and-headphones/c:29957/'
```
```
INFO: Testing url http://www.argos.co.uk/browse/technology/ipod-mp3-and-headphones/c:29957/ run 1
[2018-03-24 17:23:43] ERROR: Error: Command failed: ffprobe -f lavfi -i movie='/sitespeed.io/sitespeed-result/www.argos.co.uk/2018-03-24-17-23-28/pages/www.argos.co.uk/browse/technology/ipod-mp3-and-headphones/c:29957/data/video/0.mp4',signalstats -show_entries frame_tags=lavfi.signalstats.SATAVG -v quiet -print_format json```

The file being passed to `ffprobe` hasn't had the URI path removed.

PR #1961, opened 2018-03-24
Sanitize path segments when creating folder

rsiddle:
This should fix issue #1959. Some URI paths contain characters they probably shouldn't, which can break directories. It would make sense to santize them with `a-z, 0-9, _-.`to start with. Perhaps a I18n sanitizer may be better in the long term though?


soulgalore:
Hi @rsiddle thanks I missed this PR, let me check it later today. 
Best
Peter

soulgalore:
Looks good to me but want @tobli input when we has time for how we want to solve this in the long run!

PR #1988, opened 2018-04-19
tests: use npm ci

DanielRuf:
`npm ci` is much faster and is meant for CI. This PR is meant for evaluating the new ci command and testig the performance and compare the logs of Travis CI.

https://docs.npmjs.com/cli/ci
http://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,10 @@ addons:
     - g++-4.8
     - google-chrome-stable
   firefox: 'latest-nightly'
+install:
+  - npm ci
 before_install:
+- npm i -g npm@latest
 - firefox --version 2>/dev/null
 - google-chrome --product-version
 - python --version
--- a/package-lock.json
+++ b/package-lock.json
@@ -5829,7 +5829,7 @@
     },
     "onetime": {
       "version": "1.1.0",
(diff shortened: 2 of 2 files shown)


soulgalore:
Hi @DanielRuf ah cool, didn't know about. The failing is because of the private WPT key, we need to fix that but I'm not sure about the best. We had some tries but luck (like PRs shouldn't run the WebPageTest tests).

Best
Peter

tobli:
Hi @DanielRuf,
Try rebasing your PR against master, there's a change to the Travis config that hopefully fixes the encrypted key issue for PR builds.

tobli:
Great that the tests are back to green. The reason why npm ci is not faster is that we cache the node_modules folder in Travis. That said, I'm open to reasons why we should perhaps use npm ci anyway, even if it would be slightly slower.

DanielRuf:
You are right and node_modules is cached already (even if it is not cached with yarn). So I think we can close this PR as the benefits are not so big compared to the current setup.

In general `npm ci` uses directly the dependency tree from the lockfile and recreates it (the lockfile is the source of truth and there are other CI specific steps done).

PR #1990, opened 2018-04-20
Headless will not work with video/visualMetrics, someone will try so let us be one step ahead

soulgalore:


lib/cli/cli.js
```diff
@@ -15,6 +15,13 @@ let yargs = require('yargs'),
   webPageTestConfig = require('../plugins/webpagetest/index').config,
   slackConfig = require('../plugins/slack/index').config;
 
+function validateInput(argv) {
+  if (argv.headless && (argv.video || argv.visualMetrcs)) {
```
tobli: visualMetrics is misspelled here.

Issue #777, opened 2015-11-16
is there any way to send metrics to StatsD?


ewe-biwu:
Hi,

Our team wants to use SiteSpeed.IO to run some web tests, and build dashboard in Grafana for page performance analysis.  From sitespeed.io documentation, https://www.sitespeed.io/documentation/performance-dashboard/#graphite, it seems like data will be directly sent to Graphite, but our team currently mainly support this flow: data from application -> statsD -> Graphite-> grafana, and it might not work if data directly goes to Graphite.  i'm  wondering, if there is any way SiteSpeed.IO can send data to StatsD and then from our end to pump data to Graphite?

Thanks!
Bingbing

soulgalore:
Closing old 3.x issues. Please reopen if you this still is an issue for 4.0.

scabenga:
@soulgalore, just checking if this was ever resolved.  I have the same setup as @ewe-biwu describes above, however when sending the metrics to statsd, nothing shows in Graphite.

From what i can tell is that the syntax for statsd is slightly different.

Graphite:
`echo "metric_name metric_value `date +%s` | nc -q0 127.0.0.1 2003`

StatsD
`echo "metric_name:metric_value|type_specification" | nc -u -w0 127.0.0.1 8125`

soulgalore:
@scabenga no it will not, it is as you said. If we want StatsD to work we need PR that makes it configurable.

PR #1994, opened 2018-04-23
Add StatsD support (with bulking)

omrilotan:
I'm optimistic about the comment in issue #777:
> ❝  If we want StatsD to work we need PR that makes it configurable. ❞

### StatsD Support
Send graphite data via StatsD, Lintening on UDP connections (DGRAM) via `graphite.statsd` [boolean]

### Limit bulk sizes
Avoid `EMSGSIZE` errors from services limiting data size (like default StatsD) by setting a `graphite.bulkSize` [?number]

## Technical implementation
I have created a graphite sender superclass, supporting bulkSize feature on both default Graphite sender and the new StatsD sender.

The plugin's index (select sender) and data generator (determine format) use an explicit check (`helpers/is-statsd` module), `graphite.statsd` option must equal exactly to `true`.

## examples

| Existing settings (no statsd and no bulk) | Setting statsd to true and bulkSize to 20
| --- | ---
|  |

### DIFF
--- a/docs/documentation/sitespeed.io/configuration/config.md
+++ b/docs/documentation/sitespeed.io/configuration/config.md
@@ -51,6 +51,8 @@ Graphite
   --graphite.namespace           The namespace key added to all captured metrics.  [default: "sitespeed_io.default"]
   --graphite.includeQueryParams  Whether to include query parameters from the URL in the Graphite keys or not  [boolean] [default: false]
   --graphite.arrayTags           Send the tags as Array or a String. In Graphite 1.0 the tags is a array. Before a String  [boolean] [default: true]
+  --graphite.statsd              Uses the StatsD interface [boolean] [default: false]
+  --graphite.bulkSize            Break up number of metrics to send with each request. Null means send everything at once [number] [default: null]
 
 Plugins
   --plugins.list    List all configured plugins in the log.  [boolean]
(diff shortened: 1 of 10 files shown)


soulgalore:
Hi @omrilotan ah super cool, let us have a look the coming days! One small thing that isn't obvious from the docs is that _docs/documentation/sitespeed.io/configuration/config.md_ is automatically generated when we do a release (that way we can merge things and then push the new release and update the docs at the same time), so if you can remove that change that will be perfect. It will automatically change when we push the release.

Also just FYI: master is now the coming 7.0 so it will not be released until ... before the summer :) But let us work in the PR here first and then we can also move it to the 6.x branch so we can release it.

Best
Peter

omrilotan:
I'll be happy to apply to 6.x.x after review

soulgalore:
Hi @omrilotan thanks for a really well worked through PR! 

The only thing for me was using inheritance, I used to work a lot with Java a long time ago and that made me allergic to inheritance, but I guess that is a problem on my side only :)

Thanks again!

omrilotan:
Thank you for your quick responses, @soulgalore, should I introduce this change to another branch to include it in 6.x.x?

soulgalore:
For me it is ok to release it in 7 (before summer whenever that happens), but if you feel like it, it would be no problem to add it to 6.x. Thanks again @omrilotan !

PR #1997, opened 2018-04-25
📖 Update docs: Browsertime's percentile and decimal options

omrilotan:
When I needed to reduce the number of metrics sending to Graphite, I've seen the option to apply by own set of percentiles for browsertime that was not documented. This is the documentation for it.


omrilotan:
Just saw there's duplication and discrepancy between the options documentation here and in browsertime. I'll try to create a more unified solution and suggest that

PR #2007, opened 2018-04-29
Fix html render error from introducing ’iteration’

tobli:


soulgalore:
Thanks @tobli I've missed checking for errors and then the kids got home and I missed looking for errors.

PR #2013, opened 2018-05-03
New S3 plugin

soulgalore:
Skip the old and outdated S3 plugin and use our own instead where you can configure
almost everything for your S3 instance/upload.

### DIFF
--- a/lib/plugins/s3/index.js
+++ b/lib/plugins/s3/index.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const fs = require('fs');
+const fs = require('fs-extra');
 const AWS = require('aws-sdk');
 const readdir = require('recursive-readdir');
 const pLimit = require('p-limit');
--- a/lib/plugins/s3/index.js
+++ b/lib/plugins/s3/index.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const fs = require('fs-extra');
+const path = require('path');
 const AWS = require('aws-sdk');
 const readdir = require('recursive-readdir');
 const pLimit = require('p-limit');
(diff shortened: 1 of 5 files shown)


lib/plugins/s3/index.js
```diff
@@ -1,109 +1,128 @@
...
+        resolve();
+      }
+    };
+    const options = { partSize: 10 * 1024 * 1024, queueSize: 1 };
+    const params = {
+      Body: stream,
+      Bucket: s3Options.bucketname,
+      ContentType: contentType,
+      Prefix: s3Options.path || prefix,
+      Key: prefix + file.replace(baseDir, ''), // TODO is this correct if we set the path?
```
tobli: I think the key should be:
```
// path of the file inside of the dir
const subPath = path.relative(baseDir, file);
let key;
if (s3Options.path) {
  key = path.join(s3Options.path, subPath);
} else {
  key = path.join(prefix, subPath)
}
```

tobli: or ```const key = path.join(s3Options.path || prefix, subPath)```

lib/plugins/s3/index.js
```diff
@@ -1,109 +1,128 @@
...
+      } else {
+        resolve();
+      }
+    };
+    const options = { partSize: 10 * 1024 * 1024, queueSize: 1 };
+    const params = {
+      Body: stream,
+      Bucket: s3Options.bucketname,
+      ContentType: contentType,
+      Prefix: s3Options.path || prefix,
```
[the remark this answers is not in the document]
soulgalore: Hmm this is what we used in the old plugin, let me dig into it.

lib/plugins/s3/index.js
soulgalore: Need to test it.

lib/plugins/s3/index.js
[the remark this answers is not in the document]
soulgalore: You are right, removed it.

lib/plugins/s3/index.js
soulgalore: Fixed.

PR #2014, opened 2018-05-03
FIX statsd timer

omrilotan:
My implementation was error
StatsD supports the types:

* counter `c` (default)
* timer `ms`
* set `s`
* gauge `g`
* histogram `h`

So `t` defaulted to counter


soulgalore:
Ah thanks @omrilotan ! Looks like there's one more test that needs to be changed too.

omrilotan:
Yeah, thanks, @soulgalore . Sorry about introducing this bug.
After this one goes thru I'll create an update PR for 6.x, too

PR #2015, opened 2018-05-03
FIX statsd timer

omrilotan:
Apply fix from https://github.com/sitespeedio/sitespeed.io/pull/2014

> My implementation was error
> StatsD supports the types:
> 
> * counter `c` (default)
> * timer `ms`
> * set `s`
> * gauge `g`
> * histogram `h`
> 
> So `t` defaulted to counter


soulgalore:
Thanks again @omrilotan

PR #2028, opened 2018-05-11
influx db sets regexp tag group

amic81:
add option --influxdb.regexp.tag.group

Sets the regular expression that identifies the characters that will be replaced by the _ character in the tag group


soulgalore:
Hi @amic81 cool thanks! There are some lint issues: You can fix that by `npm run lint:fix`and then push again! Also if you can exclude the change in  `docs/documentation/sitespeed.io/configuration/config.md` - that files is auto-generated when we do a new release (need to find a way so that is easier to understand).

Thanks again!
Peter

soulgalore:
@amic81 I see now, I maybe simplified it too much: You can choose --influxdb.groupSeparator character so you could add . but it isn't completely correct since it will change every unsafe character to that char ... hmm.  So I see your point by adding a full regex ...

Issue #2024, opened 2018-05-08
InfluxDB annotation tags


Icecold777:
Hi there,

Could you please explain why do you append (actually rewrite) category influxdb tag in annotations only? 
That means it's not possible to add manually submitted tags with --influxdb.tags

soulgalore:
Hi @Icecold777 
I'm not following completely. Do you mean that when you add --influxdb.tags they aren't added to the  metrics at all until you add annotations? Can show us how you run (what parameters you use)? And also if you can show the result so I understand what's wrong :)

Best
Peter

Icecold777:
Hi,
I mean that tags added with --influxdb.tags are not added to annotations at all.
When I use _--influxdb.tags category=APAC,build=2.12.2_ category tag is updated while build tag don't appear in events table and appear in all other tables.

it happens because of following lines in send-annotation.js 
const keyAndValue = row.split('=');
if (keyAndValue[0] === 'category')tags += `,${keyAndValue[1]}`;

soulgalore:
Aha I see, we miss out on them. Is it possible for you to do a PR to master (and then I can cherry pick and also add it to the current 6.x) and I can release it?

soulgalore:
@Icecold777 that was perfect thanks a lot!

Icecold777:
Hi, @soulgalore. I see it's merged to master but when I install sitespeed.io with npm or pull the latest Docker image the code is not changed ( there is no fix). Could you clarify this please?

soulgalore:
Hi @Icecold777 **latest** we try to keep at the latest stable release. We haven't released any stable since the merge. Master in git is the coming 7.0-

If you wanna try what's in master, use the Docker tag 7.0.0-alpha.4.

With npm it is also an alpha tag, so you need to specify that tag when you install.

Best
Peter

soulgalore:
This is released in 7.0

PR #2031, opened 2018-05-11
update send-annotations.js to add tags to InfluxDB events table

Icecold777:
**Description**

This PR is related to #2024. After fix all InfluxDB tags added with --influxdb.tags would appear in events table and be used in annotations at Grafana dashboard.


soulgalore:
Thanks @Icecold777 , doing a alpha release today or tomorrow.

PR #2034, opened 2018-05-12
Fix tab spacing for configuration as JSON example

benedictdudel:
Replace tabs with spaces in the example for the configuration as JSON


soulgalore:
thanks @benedictdudel :)

PR #2079, opened 2018-06-13
choose adb port

soulgalore:


soulgalore:
This isn't working as expected so we shouldn't merge that.

PR #2096, opened 2018-07-11
Firefox sometimes creates a HAR that is broken, guard PageXray against that.

soulgalore:
The HAR can in some cases have a content size that is null from Firefox and
that breaks the spec and PageXray stats. Let us guard against that until
the upstream issue is fixed. #2097


soulgalore:
@tobli is there a better way to check or?

soulgalore:
Merging this so it can go out with tomorrows release.

Issue #2106, opened 2018-07-23
statsd weird formatting 


marceloboeira:
I want to push sitespeed.io to datadog, which uses mostly the standard statsd protocol.

I've tried to use the config as:

```json
{
    "browsertime": {
        "connectivity": {
            "profile": "3gfast"
        },
        "video": false,
        "visualMetrics": true,
        "iterations": 5,
        "chrome": {
            "timeline": true
        }
    },
    "chrome": {
        "timeline": true
    },
    "crawler": {
        "depth": 1
    },
    "plugins": {
        "remove": ["screenshot"]
    },
    "graphite": {
        "statsd": true,
        "bulkSize": 20,
        "host": "datadog-statsd.kube-system",
        "port": "8125"
    },
    "utc": true,
    "mobile": true,
    "video": false,
    "firstParty": ".hey.car"
}
```

And it didn't work. So, I've tried to push to my localhost and monitor the statsd port with netcat:

```
root@marcelo-sitespeed-55fbb56dd-x8dxg:/sitespeed.io# nc -ulp 8125
sitespeed_io.default.pageSummary.hey_car._.chrome.3gfast.browsertime.statistics.cpu.events.XHRReadyStateChange.p90:1|ms 1532331394
sitespeed_io.default.pageSummary.hey_car._.chrome.3gfast.browsertime.statistics.cpu.events.XHRReadyStateChange.p99:1|ms 1532331394
sitespeed_io.default.pageSummary.hey_car._.chrome.3gfast.browsertime.statistics.cpu.events.XHRReadyStateChange.max:1|ms 1532331394
sitespeed_io.default.pageSummary.hey_car._.chrome.3gfast.browsertime.statistics.cpu.events.MajorGC.median:6|ms 1532331394
sitespeed_io.default.pageSummary.hey_car._.chrome.3gfast.browsertime.statistics.cpu.events.MajorGC.mean:6|ms 1532331394
sitespeed_io.default.pageSummary.hey_car._.chrome.3gfast.browsertime.statistics.cpu.events.MajorGC.mdev:0|ms 1532331394
sitespeed_io.default.pageSummary.hey_car._.chrome.3gfast.browsertime.statistics.cpu.events.MajorGC.min:6|ms 1532331394
sitespeed_io.default.pageSummary.hey_car._.chrome.3gfast.browsertime.statistics.cpu.events.MajorGC.p10:6|ms 1532331394
[...]

PR #2107, opened 2018-07-23
[FIX] StatsD entry format

omrilotan:
Addressing issue https://github.com/sitespeedio/sitespeed.io/issues/2106 : StatsD formatting with excess of timestamp at end of line
```
sitespeed_io.default.pageSummary.hey_car._.chrome.3gfast.browsertime.statistics.cpu.events.XHRReadyStateChange.p90:1|ms 1532331394
```

Some implementations of StatsD will allow this excess while other do noe. None of them use it.
`util`'s string formatter places any "extra" arguments at the end of the string - and this is why this string was included while not needed by the format.
I have changed the number of arguments to exclude this string from every line. This should fix the issue

### DIFF
--- a/lib/plugins/graphite/data-generator.js
+++ b/lib/plugins/graphite/data-generator.js
@@ -7,6 +7,9 @@ const flatten = require('../../support/flattenMessage'),
   formatEntry = require('./helpers/format-entry'),
   isStatsd = require('./helpers/is-statsd');
 
+const STATSD = 'statsd';
+const GRAPHITE = 'graphite';
+
 function keyPathFromMessage(message, options, includeQueryParams) {
   let typeParts = message.type.split('.');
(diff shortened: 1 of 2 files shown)


marceloboeira:
@omegahm thanks for the fix 👍

omrilotan:
This solution seems to be working
image
After approved by owner, I can add a backwards compatible fix for 6.x

soulgalore:
Thanks @omrilotan for taking the time to fix this, really appreciate it! I'm releasing ASAP.

soulgalore:
Will let it roll out first on dashboard.sitespeed.io so it will happen tonight in about 5-6 hours.

marceloboeira:
I've tested on prod here, building from the branch of the PR. Its pushing data to datadog (statsd server). Thanks!

PR #2110, opened 2018-07-23
[FIX] StatsD entry format (#2107 clone)

omrilotan:
Applying fix #2107 to version 6 (where StatsD support was introduced)


soulgalore:
Hey @omrilotan do you still use 6.x? When we pushed 7.0 2018-05-24 we stopped updating the 6.x branch, so there are new browser updates etc in that Docker container.

Best
Peter

omrilotan:
@soulgalore No, I use the autobuild. If V6 is not supported, I'll close this pull request. Thanks 👌

Issue #2127, opened 2018-08-07
Add possibility to add alias for URL from the command line


soulgalore:
This will also make it possible to use the functionality directly from a config.json.

soulgalore:
Merged an coming in 7.5

PR #2133, opened 2018-08-08
Use alias from cli #2127

soulgalore:
@beenanner want your feedback on this approach!


shiftie:
would be a nice feature!

soulgalore:
This needs to be changed before merged, so let us do that after 7.5. 
Right now the URL is used as alias if you supply alias to one/ a couple of URLs but not to all of them. That is bad because it will change the key in Graphite.

I think we should revert (and fix) the first implementation:  check that the user use the same amount of aliases as URLs.

PR #2135, opened 2018-08-10
Collect content/transfer size per domain

soulgalore:
Collect transfer/content size per domain. Next step is to find a way to make it fit in the HTML result table.


soulgalore:
Lets redo this if there's an actual need for it.

PR #2144, opened 2018-09-05
Include screenshots in annotations

soulgalore:
If we include screenshots it makes it easier to know what's going on:
screen shot 2018-09-05 at 12 22 46


soulgalore:
Want to fine tune this a couple of iterations before we merge.

Issue #2173, opened 2018-10-05
Domain name on Influxdb


quartex:
Hi,
we are using Influxdb to save scan report data and we have a problem on domain name because the saved value for domain is in IDN format (example edonè.com became xn--edon-8oa.com) and this is a problem for a Db to have a right order or search not using specific function in where clause. 
The last db (Influxdb e Graphite too) can save the plain format without a problem.

Is it a bug or a or something desired for other motivation ?

Thank you very much

soulgalore:
It needs to be behind a switch/cli flag so it is backward compatible. One way could be to just fix it for Graphite/Grafana probably somewhere in https://github.com/sitespeedio/sitespeed.io/blob/master/lib/support/tsdbUtil.js

quartex:
Ok ... we try to find best solution and we want to use punycode.js library to fix it 
Is it ok ?

mendaomn:
Hey,

an idea would be to simply call `punycode.toUnicode(idn_encoded_url)` (or something similar), but I'm struggling to find where that line may go. Where is the "wrongly formatted" url in [that](https://github.com/sitespeedio/sitespeed.io/blob/master/lib/support/tsdbUtil.js#L19)? 

Do you have any ideas?

soulgalore:
Hey @mendaomn it's the group. The idea from the beginning was to call it "group" to be able to group URLs on other ways that domains but in practice we only group by domain so group is domain.

Best
Peter

mendaomn:
Alright, I'll put together a PR (:

PR #2177, opened 2018-10-10
Add punycode to escape characters and convert the group to unicode

mendaomn:
Hi,

as discussed in #2173, I added punycode to convert group to unicode

Please let me know if I can do anything to improve this


soulgalore:
Hi @mendaomn thanks for the PR! I'll have a look this weekend.

Best
Peter

soulgalore:
Sorry for being so late with this. I want to avoid adding more dependencies, but let me have a look the coming weeks.

PR #2182, opened 2018-10-16
Fix Grafana screenshots

soulgalore:


julrich:
Thanks for the quick fix! :)

PR #2285, opened 2019-02-02
Fixes #2072 by adding endpoint as an configurable argument for s3 options

### LINKED ISSUE #2072: Test out s3 plugin compatibility with DO Spaces
It looks like we could possible add documentation around the use of DO spaces since they are compatible with the S3 API. This would be good to get our users another option for html storage.

https://www.digitalocean.com/products/spaces/

beenanner:
This should allow #2072 to work with Digital Ocean spaces
Example config:
```

    "s3": {
      "endpoint": "sfo2.digitaloceanspaces.com",
      "key": "xxxxxxxxx",
      "secret": "xxxxxxxxxx",
      "bucketname": "some-bucket-for-th-results",
      "removeLocalResult": true,
      "acl": "public-read"
   },
```


soulgalore:
Great @beenanner I'll merge this and then when you have time please update the docs. I've just pushed latest of Browsertime with some scripting updates that need doc updates so will fix that early tomorrow for a new release.

PR #2319, opened 2019-02-17
Add support for crawler exclude patterns

### LINKED ISSUE #1929: Allow to exclude URLs from crawling
It would be great to have a feature to exclude URL paths from crawling. I would like to analyse `http://www.example.com/path1/*` but not `http://www.example.com/path2/*`

Maybe this can help: <https://github.com/simplecrawler/simplecrawler/wiki/Target-crawling-desired-resources-and-ignore-others>

fholzer:

### Description
Add support for crawler exclude patterns. Multiple patterns may be provided. Resolves #1929. 

At the moment the patterns are matched against the full URL. Feedback is appreciated. Will squash and add documentation if you're happy with this PR


lib/cli/cli.js
```diff
@@ -983,6 +988,15 @@ module.exports.parseCommandLine = function parseCommandLine() {
         );
       }
     })
+    .coerce('crawler', crawler => {
+      if (crawler.exclude) {
+        if (!Array.isArray(crawler.exclude)) {
+          crawler.exclude = [crawler.exclude];
+        }
+        crawler.exclude = crawler.exclude.map(e => new RegExp(e));
```
fholzer: Wasn't sure where to put this. At least here it would fail early in case the user provided an invalid regular expression.

lib/plugins/crawler/index.js
```diff
@@ -68,7 +89,7 @@ module.exports = {
 
             if (pageCount >= maxPages) {
               log.info('Crawler stopped after %d urls', pageCount);
-              crawler.stop();
+              crawler.stop(true);
```
fholzer: Even before my changes, when specifying e.g. `--crawler.maxPages 5` most of the times I would get 5 pages in the report, but sometime I'd get 6. I guess that's because the crawler already discovered the 6th URL by the time `crawler.stop()` was called. I hope adding `true` will fix that. See https://github.com/simplecrawler/simplecrawler/blob/be994f01bc2b055aabb192290babe647f7a08ca1/lib/crawler.js#L1829

If that's not good enough, an additional pageCount check needs to be added at the top of the  `crawler.on('fetchcomplete', ...)` callback, I guess.

soulgalore:
Hi @fholzer wow so nice you spend time to fix this! I've been using the crawler so much the last 3-4 years, so I haven't given it much love. 

Thanks for the PR!

PR #2332, opened 2019-02-21
Bugfix/annotations extra message

mliljedahl:
### Description

When sending annotations with `--graphite.annotationMessage` and not providing a screenshot the message and the extraMessage gets concatenated without any space.

This commit fixes that and follows the same standard that is used when having a screenshot; separating the messages with ` - `.


soulgalore:
Thank you @mliljedahl !

PR #2355, opened 2019-03-01
cpu throttle to run inside docker

kabilanravi:

### Description
This change is for cpu throttling inside docker containers. Reference for cpulimit functionality is in "http://cpulimit.sourceforge.net". 

Usage :- "docker run --shm-size=1g --rm -v "$(pwd)":/sitespeed.io sitespeedio/sitespeed.io https://www.sitespeed.io/ -t 100"
CPU throttle number behaves exactly as given in the cpulimit instruction documentation

### DIFF
--- a/docker/scripts/cpu-throttle.sh
+++ b/docker/scripts/cpu-throttle.sh
@@ -1,0 +1,10 @@
+while true; do
+    if ps -ef | grep driver | grep use-mock-keychain > /dev/null
+    then
+        PROCESS_PID=$(ps -ef | grep driver | grep use-mock-keychain | awk '{print $2}')
+        echo "Running"
+        echo $PROCESS_PID
+        cpulimit -p $PROCESS_PID -l $1
+    fi
+    sleep 1
+done
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,9 @@ RUN sudo apt-get update && sudo apt-get install libnss3-tools \
  mkdir -p $HOME/.pki/nssdb && \
  certutil -d $HOME/.pki/nssdb -N
 
+# For cpu throttling
+RUN sudo apt-get install cpulimit
+
 ENV PATH="/usr/local/bin:${PATH}"
 
 RUN wpr installroot --https_cert_file /webpagereplay/certs/wpr_cert.pem --https_key_file /webpagereplay/certs/wpr_key.pem
@@ -30,6 +33,9 @@ RUN npm install --production
 COPY . /usr/src/app
 
 COPY docker/scripts/start.sh /start.sh
+
+# Script for cpu throttling
+COPY docker/scripts/cpu-throttle.sh /cpu-throttle.sh
 
 ## This is to avoid click the OK button
 RUN mkdir -m 0750 /root/.android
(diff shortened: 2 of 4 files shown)


docker/scripts/cpu-throttle.sh
```diff
@@ -0,0 +1,10 @@
+while true; do
+    if ps -ef | grep driver | grep use-mock-keychain > /dev/null
+    then
+        PROCESS_PID=$(ps -ef | grep driver | grep use-mock-keychain | awk '{print $2}')
```
soulgalore: Can you share a little with some comments how this works? What process id:s are finding/searching?

soulgalore:
Great thanks a lot @kabilanravi ! 

I just want to structure it a little other way, I can do that if I have write access to the PR or I can merge and do the change, so I don't take up your time. Wanna change so we use a Docker variable to turn on the throttling (like in https://github.com/sitespeedio/sitespeed.io/blob/master/docker/scripts/start.sh#L20) because then we have the same way of handling Docker specific code all over the place). Then we can also move the install to the Docker container that includes the browsers, so Browsertime users also can use the same.

Thanks again @kabilanravi ! Let me know if you from your tests got any experience from what numbers and what kind of instances you have been running on.

docker/scripts/cpu-throttle.sh
kabilanravi: `ps -ef | grep driver | grep use-mock-keychain
PROCESS_PID=$(ps -ef | grep driver | grep use-mock-keychain | awk '{print $2}')` This finds the pid of the selenium driver instance to which we are going to limit the cpu. For every run pid varies, because we kill the process and start it again. For continuous throttling, this script will look for pid change and throttle it immediately.

kabilanravi:
Thanks @soulgalore for considering my changes. 

I am running sitespeed with cpu throttling for almost 2 months now. The idea is, we have network throttling in sitespeed. But I have a requirement where we need to test our product in low cpu machine as well. So without using 2 different configuration machine (docker), we use 2 run commands. One for full cpu speed, another for 50% throttled speed. Using the throttled cpu results we optimised our product and saw better results in visualcomplete metrics.

I have added you as collaborator in the forked repo.

soulgalore:
I'll have a look this weekend, thinking maybe it would be cleaner to use [cgroups](https://en.wikipedia.org/wiki/Cgroups) and integrate the start/stop into Browsertime directly.

soulgalore:
Started to look at this last night, I think it will take a while to get it up and running. I started to test yesterday if I could get stable metrics but it was hard. I would like to test out [cgroups](https://www.digitalocean.com/community/tutorials/how-to-limit-resources-using-cgroups-on-centos-6) but that would take some time.

kabilanravi:
cgroups should do the job. I was implemented this as we need a quick solution. I will take a look at cgroups. Will update you soon. Thanks @soulgalore

Spunkie:
Any updates on this? It will be super useful to be able to throttle CPU usage. That way sitespeed.io can run on small/underpowered servers while still providing consistent rendering times.

soulgalore:
@Spunkie If you use Chrome, checkout CPUThrottlingRate, that is what I use for testing.

Issue #2329, opened 2019-02-21
Google Cloud storage bucket


mliljedahl:
I just wrote a plugin for storing the results in a Google Cloud storage bucket. But before I make a pull request of it I would like to align it to the S3 plugin that already exists. Then I came across a few questions. 

When trying to implement the `path` option that exists in S3 and also sending annotations to Graphite the resultUrls is completely ignoring the `path` option giving me:

```INFO: [sitespeedio] Find the result at https://storage.googleapis.com/bucketname/www.my-domain.com/2019-02-21-10-49-01```

Instead of (setting path to `my-sub-folder/test1`):

```INFO: [sitespeedio] Find the result at https://storage.googleapis.com/bucketname/my-sub-folder/test1```

This gives wrong urls in the annotations.
 
I do not have the possibility to test the S3 storage plugin but to me it looks like the same problem exists there? I have also inspected the `message` in the Graphite plugin and cant find any info about the path in there and also the Graphite plugin looks like it is run before storing the results in S3?

As a bonus it would be nice to not having to specify a `--resultBaseURL` when using S3 (or any other storage) where it would be possible to default to an url connected to the storage provider; `https://storage.googleapis.com` in the GCS case if it is not set.

Tested using the master branch pulled a few minutes ago.

soulgalore:
Hi @mliljedahl aha cool!

> I do not have the possibility to test the S3 storage plugin but to me it looks like the same problem exists there? I have also inspected the `message` in the Graphite plugin and cant find any info about the path in there and also the Graphite plugin looks like it is run before storing the results in S3?

Hmm yeah that sounds like a bug and something we missed then. I haven't used that part for a long time and I'm not 100% sure how it would work, need to look into it.

Best
Peter

soulgalore:
> As a bonus it would be nice to not having to specify a `--resultBaseURL` when using S3 (or any other storage) where it would be possible to default to an url connected to the storage provider; `https://storage.googleapis.com` in the GCS case if it is not set.

Yes that would be cool, good idea, let us add that as separate issue!

soulgalore:
Ping @beenanner since I guess you are also interested in this :)

beenanner:
Very much so let me take a look at this!

soulgalore:
@beenanner cool! Do you mean you help @mliljedahl if he needs any help or do you mean you fix the bug #2330 he found? Or implement #2331? :)

beenanner:
[...]

PR #2360, opened 2019-03-06
Feature/google cloud storage plugin

mliljedahl:
### Description
This PR implements support for Google Cloud Storage (GCS) buckets, as an alternative to the AWS S3 buckets. Documentation is in a separate PR. Issue number for this one is #2329.


soulgalore:
@mliljedahl I think Travis is failing because of that the npm-shrinkwrap.json doesn't have the items from your fix in package.json, so if you can include an updated  npm-shrinkwrap.json in the PR that would be super (it isn't super clean, wonder if there's a way of doing that in a better way).

lib/gcs/index.js
```diff
@@ -0,0 +1,93 @@
...
+    validation: 'crc32c',
+    gzip: true,
+    metadata: {
+      metadata: {
+        cacheControl: 'public, max-age=31536000'
+      }
+    }
+  };
+
+  await bucket.upload(file, params);
```
soulgalore: Small thing: here you can just return the promise and the user of the function will choose to wait/then etc.

```suggestion
  return bucket.upload(file, params);
```

PR #2368, opened 2019-03-07
Fix/gcs move to plugins folder

mliljedahl:
### Description
Sorry for this, the GCS plugin was placed in the wrong folder in previous commit. This fixes that.


soulgalore:
I've should have spotted that when I merged, sorry!

PR #2397, opened 2019-03-28
Improve use of browsertime.connectivity.alias option

fholzer:
- [X] Update the documentation https://github.com/sitespeedio/sitespeed.io/tree/master/docs in this PR

### Description
Browsertime supports `connectivity.alias` option to give connectivity profiles a custom name, though this is not documented in the Sitespeed cli help, and not consistently used in Sitespeed. (only used for writing to DB, but not in reports)

This PR
- adds option to sitespeed cli
- uses the alias in html reports
- adds documentation

Cheers,
Ferdinand


soulgalore:
Ah great, thanks @fholzer ! Can you just remove the change in config.md, that file is auto-generated on a release, so when we release your change, the documentation will automatically be updated.

Best
Peter

fholzer:
oh, didn't know. Done
Cheers/f

soulgalore:
Perfect thanks @fholzer !

PR #2398, opened 2019-03-28
Update docs with missing configuration option

morty29:
### Description
Added one missing option


soulgalore:
Ah thanks @morty29 I've missed that. One thing: the config.md is auto updated when we do a release from https://github.com/sitespeedio/sitespeed.io/blob/master/lib/cli/cli.js so when we run the release script we generate that file (see https://github.com/sitespeedio/sitespeed.io/blob/master/release.sh#L32).

So we need tio update the CLI file, then it config.md will be fixed (I'm planning for a release on Saturday this week so then it will be fixed).

Best
Peter

PR #2438, opened 2019-04-26
Fixes/gcs fixes

mliljedahl:
### Description
Quite critical fixes for the GCS plugin when storing the result for multiple runs at the same time and the connections becomes to many. Also includes fix for when using the cloud bucket for hosting the results as a web site. More info in each commit message.


soulgalore:
Thank you @mliljedahl I'll push it tonight.

Issue #1277, opened 2016-10-28
Add specific documentation for sending timings per asset/domain


soulgalore:
To make it easier to catch problems with third parties.

https://github.com/sitespeedio/grafana-bootstrap-docker/issues/2

jpvincent:
If you need to update the documentation, I can provide what I did.
First, enable more data to be sent to graphite, for example in a config.json file given when running sitespeed.io
```json
  "metrics.filter":[
    "webpagetest.pageSummary.data.median.firstView.domains.*.bytes",
    "webpagetest.pageSummary.data.median.firstView.domains.*.requests"
  ],
```

Then create a dashboard. I made one for myself that lists # of bytes and requests for the top 10 domain, and show the repartition between the main domain and the others 👍  
![image](https://cloud.githubusercontent.com/assets/427625/21504617/c203cfe6-cc60-11e6-87f1-d03fcb1bddda.png)

I dont know how to provide the code though : maybe as a default graph, but with a sticky note explaining how to enable it ?

soulgalore:
Thank you @jpvincent for the update, I will add it to the docs when I have time (and give you cred). Thanks again!

soulgalore:
With the new 3rd party functionality in late 8 and 9 I think the CPU time spent for 3rd parties is the way to go. What we could do is and total time spent downloaded per 3rd party (today we only have CPU). Let me have a go at that.

soulgalore:
Made a PR, coming in next release.

soulgalore:
It has happened in four years, it will never happen.

PR #2487, opened 2019-05-30
Track 3rd party download time per tool #1277

soulgalore:


soulgalore:
Low priority so let me close it.

Issue #2673, opened 2019-10-02
Metrics on per-iteration basis


kevinlacotaco:
We are using the sitespeed + statsd interface with multiple iterations. This provides us aggregated metrics. When piping this to datadog this means every aggregate is its own metric. Datadog then also provides aggregates based on the metrics provided. 

Is it possible to get the metrics fired to graphite/statsd interface on a per-iteration basis without the aggregates. I was looking at how all the messages were passed and was wondering if it existed.

- Using docker-compose (no additional plugins besides the ones provided) with 4 URLs under test and 10 iterations per. We also upload to S3 the html reports (which the aggregates are necessary for). 
- Version: 10.0.1 (I see 10.1.1 is out)
- OS: Docker provided image, Base OS is MacOS (locally)

It looks like there is a method to the iteration messages because I turned on the 'analysisstorer' plugin to checkout the provided JSONs results and found one for each iteration. 

Example metrics:

```
<namespace>.pageSummary.<page-alias>.chrome.native.browsertime.statistics.timings.loadEventEnd.mean:<number>|ms|
<namespace>.pageSummary.<page-alias>.chrome.native.browsertime.statistics.timings.loadEventEnd.mdev:<number>|ms|
<namespace>.pageSummary.<page-alias>.chrome.native.browsertime.statistics.timings.loadEventEnd.stddev:<number>|ms|
<namespace>.pageSummary.<page-alias>.chrome.native.browsertime.statistics.timings.loadEventEnd.min:<number>|ms|
<namespace>.pageSummary.<page-alias>.chrome.native.browsertime.statistics.timings.loadEventEnd.p10:<number>|ms|
<namespace>.pageSummary.<page-alias>.chrome.native.browsertime.statistics.timings.loadEventEnd.p90:<number>|ms|
<namespace>.pageSummary.<page-alias>.chrome.native.browsertime.statistics.timings.loadEventEnd.p99:<number>|ms|
<namespace>.pageSummary.<page-alias>.chrome.native.browsertime.statistics.timings.loadEventEnd.max:<number>|ms|
```

Each one goes to DataDog and then provides a `max`, `min`, `median`. So you will have a 

```
<namespace>.pageSummary.<page-alias>.<metric>.max.max
<namespace>.pageSummary.<page-alias>.<metric>.max.95percentile
<namespace>.pageSummary.<page-alias>.<metric>.max.median
<namespace>.pageSummary.<page-alias>.<metric>.max.avg
```

[...]

Issue #2118, opened 2018-07-25
Add tag support in Graphite


soulgalore:
In the current version, we use the (old) namespace structure for metrics. We should make it configurable to also add tags.

soulgalore:
I think we will not do this since its tag or keys not keys and tags.

PR #2679, opened 2019-10-04
feat(graphite): add ability to send run metrics

kevinlacotaco:

### Description

Adding an option for graphite to send metrics on a per run basis. Defaults to the current behavior which is no iteration results. And appends the iteration number directly to the `run` key for easy determination of which run it is. 

Addresses: https://github.com/sitespeedio/sitespeed.io/issues/2673

Example metric in statsd format:
`sitespeed_io.default.run-2.www_sitespeed_io._.chrome.native.pagexray.totalDomains:1|ms`


soulgalore:
Hi @kevinlacotaco cool. 

I think we need to add a section in the docs explaining how this works. I can see people turning this on, sending them directly to Graphite with the wrong storage aggregations (https://raw.githubusercontent.com/sitespeedio/sitespeed.io/master/docker/graphite/conf/storage-schemas.conf) missing out on metrics. Can you add a section on https://www.sitespeed.io/documentation/sitespeed.io/graphite/ ?

Best
Peter

kevinlacotaco:
Yeah, I can do that. Thanks for the response.

soulgalore:
Great! Can you do that in a separate PR (I forgot to say that)? When we merge, the docs are automatically pushed. If it's two PRs I can merge this first, let it run on our test server and then merge the docs on the release, thanks!

kevinlacotaco:
Yeah! Sweet. That sounds good. Do you want some tests around this? I can add those here before as well.

soulgalore:
Yeah cool, if you find a way to do it. Normally we rely heavily on the tests on Travis but for Graphite we don't have a clear path.

kevinlacotaco:
Good to know! Thanks for the quick response.

soulgalore:
Great, I'll go through and test it this weekend! Thank you for the PR!!!

kevinlacotaco:
Thank you! Really enjoy the tool and the information it provides!

lib/plugins/graphite/data-generator.js
```diff
@@ -50,6 +50,14 @@ function keyPathFromMessage(message, options, includeQueryParams, alias) {
     // add the group of the summary message
     typeParts.splice(1, 0, graphiteUtil.toSafeKey(message.group));
   }
+  // if we are sending run metrics and we have an iteration
+  if (
+    options.perIteration &&
+    message.type.endsWith('run') &&
+    message.iteration
+  ) {
+    typeParts.splice(0, 1, 'run-' + message.iteration);
```
soulgalore: I'm not 100% sure how we should handle this in Graphite/Grafana? How do you aggregate this in your DataDog/statsd example? I'm thinking that is like now (with run-iteration) the retention policy doesn't matter because each metric will have their own key. But if just call it run (leaving out the iteration) retention matter + it's also easier to do you own aggregate. 

With the run the aggregate you will aggregate per run right, so that you rill get aggregates per first run, instead of per URL?

kevinlacotaco: We have a statsd proxy in between the sitespeed and the dogstatsd container and can modify metrics to add additional tags that are a bit more dynamic. I was thinking of moving the entire `run-<number>` to a tag to the metric. This way we can use [template variables](https://docs.datadoghq.com/graphing/dashboards/template_variables/) to create filterable dashboards based on page and others. 

We are basically trying to reduce the specificity of the metric itself and exchange that to tags. Looking at [metric overview](https://docs.datadoghq.com/graphing/metrics/introduction/#breaking-down-the-metric-query) we will use the tags in the scope and group to reduce our metric set for a certain time period.

Does that make sense? If we could add it as a tag from sitespeed that would be great as well. And then we can just call it `run`.

soulgalore: Ah I see. I think we can keep the `run-<number>` as is. I was thinking none of the standard dashboard will work with it, but you will still need do you your own for that so it doesn't matter, I was wrong :)

We haven't started using tags for Graphite yet (we have a big untouched task for it in #2118). I haven't had any experience using it yet but I see the value. Haven't thought of a good way to move that forward, if you have any ideas that much appreciated. 

I wanna do a release on Monday/Tuesday since Catalina is coming (I guess) and we have some functionality that makes testing on iphone/ipad makes it better. We can include this in that release (if we keep it the current way) or do you want to iterate if it could be a first step of adding tags?

kevinlacotaco: Awesome! I can take a look at that. That would definitely be useful.

That great news on the new functionality! I would love for this currently and we can swing back around with tag support. I am just unsure how long that may be.

lib/plugins/graphite/data-generator.js
```diff
@@ -50,6 +50,14 @@ function keyPathFromMessage(message, options, includeQueryParams, alias) {
     // add the group of the summary message
     typeParts.splice(1, 0, graphiteUtil.toSafeKey(message.group));
   }
+  // if we are sending run metrics and we have an iteration
+  if (
+    options.perIteration &&
```
soulgalore: Options here are generic options and not Graphite specific right?

soulgalore:
FYI @kevinlacotaco I've made a changed and renamed the flag #2685 to `--graphite.experimental.perIteration`- its just because I'm not 100% really sure yet of the right structure. Can you verify in master that it works as it should? 

I think there was a mixup in the first PR that if you used --graphite.perIteration the key was actually `run` and not `run-[iteration]` because there was a mixup of which options that is really used (sorry the code is a little messy), at least when I tried I got just run as the key. That's fixed now but wanna know that the structure of the data works (for you) :)

lib/plugins/graphite/data-generator.js
kevinlacotaco: Looking at the code again, I thought I had tested this out. It looks like the data-generator gets the options pre-merge and its the whole object and not graphite specific.

kevinlacotaco:
Thanks for updating that! I will definitely work on tests next.

soulgalore:
@kevinlacotaco no worries! Can you check that it works now so you can collect the metrics the way you want? Then I feel more better to release it later tonight :) But no pressure, lets add it as a experimental feature that can change whenever, and we can iterate it over the coming weeks.

kevinlacotaco:
Tested it out with `--graphite.experimental.perIteration=true` and got what I expected!

Using `https://sitespeed.io` as the example:
`sitespeed_io.default.run-1.sitespeed_io._.chrome.native.pagexray.transferSize:223010|ms`

soulgalore:
Thank you @kevinlacotaco for taking the time to test! I'll release this tonight after work.

kevinlacotaco:
Sounds fantastic! Thanks for the help and quick updates!

soulgalore:
This is now released in 10.2.0, thank you @kevinlacotaco !

vishnusn656:
Hi @soulgalore is  perIteration feature limited to Graphite? I am using InfluxDB. How can I get iteration wise metrics in influxdb?

soulgalore:
Yes it is.

PR #2681, opened 2019-10-04
docs(graphite): add config, and storage info for per iteration option

kevinlacotaco:

- [X] Update the documentation https://github.com/sitespeedio/sitespeed.io/tree/master/docs in another PR

### Description

Adding documentation from feedback in https://github.com/sitespeedio/sitespeed.io/pull/2679


kevinlacotaco:
I am a little unfamiliar with graphite as we just pipe ours to Datadog and let it handle the storage. Let me know if we should be modifying the aggregation rules too.

soulgalore:
No I don't think we should not change them. The config right now is that you can only push one metric every 10 min per key (default) meaning if you test X URLs you cannot test them more often than that (why should you hehe). But if you then switch to your new per run, and we do three runs within 10 minutes, Graphite will discard two of the runs.

kevinlacotaco:
Yeah, that Etsy guide was helpful at getting a bit up to speed. I took their example of _near real time_. Could this still discard some entries?

PR #2691, opened 2019-10-09
feat -verify http status in budget (and log errors)

thapasya-m:
on budget verify, check for responseCode from
pagexray message body and log any status code > 399

### Description
2318

### DIFF
--- a/lib/plugins/budget/verify.js
+++ b/lib/plugins/budget/verify.js
@@ -61,6 +61,10 @@ module.exports = {
           if (value && message.type === 'lighthouse.pageSummary') {
             value = value * 100; // The score in Lighthouse is 0-1
           }
+          if (value && message.type === 'pagexray.pageSummary') {
+            // count number of http server error
+            value = Object.keys(value).filter(code => code > 399).length;
+          }
           // We got a matching metric
           if (value) {
             const budgetValue = budgetForThisURL[metricType][metric];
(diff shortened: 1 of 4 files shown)


lib/plugins/budget/verify.js
```diff
@@ -55,6 +55,14 @@ module.exports = {
     // timing, request, coach etc
     for (let metricType of Object.keys(budgetForThisURL)) {
       for (let metric of Object.keys(budgetForThisURL[metricType])) {
+        // [TODO] refactor might be required
+        if (message.type === 'pagexray.pageSummary') {
+          for (let responseCode of Object.keys(message.data.responseCodes)) {
+            if (parseInt(responseCode) > 399) {
+              log.error('Error:', responseCode);
```
soulgalore: Cool, I would do a couple of changes:I would log on info level since this isn't an error in sitespeed.io and then I would also fine-tune the message: Include the URL in the message so the user knows which URL that returns the error and add a short describing text.

Then I would also. move the logging to the [PageXray plugin](https://github.com/sitespeedio/sitespeed.io/tree/master/lib/plugins/pagexray) then all users will get the information if something is wrong (you don't need to run a budget).

Then keep this here but instead of logging, push it to the budget result!

soulgalore:
Thank you @thapasya-m looked great! Can you do a small follow up PR where you just updates the docs so that the new value is there: https://www.sitespeed.io/documentation/sitespeed.io/performance-budget/#full-example 

I'll merge that when I do the next release (probably on Saturday.

thapasya-m:
> Thank you @thapasya-m looked great! Can you do a small follow up PR where you just updates the docs so that the new value is there: https://www.sitespeed.io/documentation/sitespeed.io/performance-budget/#full-example
> 
> I'll merge that when I do the next release (probably on Saturday.
Thanks, @soulgalore. I will make a follow-up documentation PR.

Issue #2705, opened 2019-10-17
Slack site summary urls with s3 not directing to index.html


kevinlacotaco:
- Using docker-compose
- Version: 10.2.0 (looking to update soon)
- OS: docker container, base OS: OSX (locally)
- Browser version: included from docker container

Slack configuration: `summary`

Using the slack configuration to get stats on our runs with s3 hooked up at a base url for summary does not include index.html. Adding `/index.html` properly loads the page.

Also, visible in the sitespeed.io slack channel: `Site summary (result)` -> points to an s3 bucket without `/index.html`

Page results do not have this issue.

kevinlacotaco:
Could be right here: https://github.com/sitespeedio/sitespeed.io/blob/master/lib/plugins/slack/summary.js#L72

soulgalore:
This was fixed and released some time ago!

PR #2707, opened 2019-10-18
Add missing index.html to Slack URL #2705

soulgalore:


lib/plugins/slack/summary.js
```diff
@@ -69,7 +69,7 @@ module.exports = function(dataCollector, errors, resultUrls, name, options) {
 
   let message = '';
   if (resultUrls.hasBaseUrl()) {
-    message = ` (<${resultUrls.reportSummaryUrl()}|result>)`;
+    message = ` (<${resultUrls.reportSummaryUrl()}index.html |result>)`;
```
kevinlacotaco: Does the summary url end in a `/'? If not, do we need to also append that?

PR #2711, opened 2019-10-22
feat: user given metric summary 

thapasya-m:
ISSUES CLOSED: 2001

### Description
2001


soulgalore:
Hi @thapasya-m wow looks super cool!

A couple of small things: 
* Add a test in `.travis.yml`so we run through the code path., You can just use one of the current tests and add the new parameter with a configuration
* Add another PR with updated docs, I think you can add a section in https://www.sitespeed.io/documentation/sitespeed.io/configuration/#the-basics so that there are two different examples.
* When we display size, can we squeeze in **(kb)** in the header? 
* I think should try to follow the the same structure of default column that we show today but we have different columns depending if you run in Docker or not so ... let us discuss if there's a clean way to fix that.

This will be great addition, thank you again @thapasya-m ! After this we should add more friendly names and then some kind of display name. 
Best
Peter

lib/cli/cli.js
```diff
@@ -1142,6 +1173,21 @@ module.exports.parseCommandLine = function parseCommandLine() {
         'Will add a link on the waterfall page, helping you to compare the HAR. The full path to your compare installation. In the format of https://compare.sitespeed.io/',
       group: 'HTML'
     })
+    .option('html.pageSummaryMetrics', {
+      type: 'array',
+      describe:
+        'Select from the list of metrics to be tested for the given URL(s)',
```
soulgalore: Can you squeeze in a full example here? Users usually have a problem understanding the difference between an array/string in the context of the cli, I've seen so many different solutions/tries :)

soulgalore:
@thapasya-m I think there's some kind of missmatch between metrics/headers in some cases. I tried using docker:
```
docker build -t sitespeedio/sitespeed.io .
docker run --rm -v "$(pwd)":/sitespeed.io sitespeedio/sitespeed.io https://www.sitespeed.io/
```

And then the page looks like this:
Screenshot 2019-10-23 at 09 05 44

soulgalore:
Else it getting close now, great work @thapasya-m :)

lib/cli/cli.js
thapasya-m: ok, Peter.

soulgalore:
Thank you @thapasya-m :)

PR #2736, opened 2019-10-30
feat: summary box on result page

thapasya-m:
ISSUES CLOSED: 2727

### Description
2727


lib/plugins/html/index.js
```diff
@@ -20,6 +20,58 @@ const defaultConfig = {
       'transferSize.css',
       'transferSize.image',
       'score.performance'
+    ],
+    summaryBoxes: [
```
soulgalore: I think you told me, but I forgot (sorry), is there any reason why we can't use the friendly names here? I know we will miss out on metrics, but maybe we just need to add more data to the friendly list (or just skip a couple of metrics)? I'm thinking it will make sense in a way that we have one way when configuring metrics.

thapasya-m: Let me try with existing friendlynames. it does have same path structure.

soulgalore:
Let us merge this and do a couple of follow up PRs!

soulgalore:
Thank you @thapasya-m  :)

PR #2820, opened 2019-12-10
Add a way to add plugins through node API

PepijnSenders:

### Description

I'm using the sitespeed.io API through a node script and wanted to add an easier way to use plugins. This way there's no need to reference them by relative path, but you can just add them to the configuration hash like so:

```js
const { run } = require('sitespeed.io');

(async () => run({ plugins: { add: [{ processMessage: noop }] } }))();
```

I'm not sure if this needs unit tests or not, so please let me know if so and I'll add them!

### DIFF
--- a/docs/documentation/sitespeed.io/configuration/config.md
+++ b/docs/documentation/sitespeed.io/configuration/config.md
@@ -123,7 +123,7 @@ Graphite
 
 Plugins
   --plugins.list    List all configured plugins in the log.  [boolean]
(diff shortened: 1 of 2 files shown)


soulgalore:
Hi @PepijnSenders thanks for your PR! I'm thinking maybe it's possible to tweak the current solution to also work with runtime? We could check if its a string or an object (with maybe some of the needed fields or whatever), then we don't need to introduce a new parameter and it will work automagical for the user. 

Best
Peter

PepijnSenders:
Good idea @soulgalore, I'll work on that!

PR #2851, opened 2020-01-21
feat(docker): add `grafana-piechart-panel` plugin

erickwilder:
This should unblock [@sitespeedio/grafana-bootstrap-docker#19](https://github.com/sitespeedio/grafana-bootstrap-docker/pull/19) as the pie chart panel
is missing from the default installation of Grafana.

- [x] ~~I'm making a big change or adding functionality so I've already opened an issue proposing the change to other contributors, so I got feedback on the idea before took the time to write precious code~~
- [x] ~~Verify that the test works by running `npm test` and test linting by `npm run lint`~~

### Description
@soulgalore [mentioned](https://github.com/sitespeedio/grafana-bootstrap-docker/pull/19#issuecomment-475890930) this was a blocked to move forward with the changes introduced in the linked pull request.

This is now shown at the Home section of Grafana:  
Screenshot 2020-01-21 at 21 59 13

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -20,6 +20,7 @@ services:
         - GF_AUTH_ANONYMOUS_ENABLED=true
         - GF_USERS_ALLOW_SIGN_UP=false
         - GF_USERS_ALLOW_ORG_CREATE=false
+        - GF_INSTALL_PLUGINS=grafana-piechart-panel
       volumes:
         - grafana:/var/lib/grafana
       restart: always


soulgalore:
Hi @erickwilder cool, I've haven't added it before because I haven't tried when you try to import a dashboard to a Grafana instance that miss the plugin (but one chart is using it), do you know?

erickwilder:
@soulgalore 

The scope of this change only affects the docker compose setup which I believe is only used when playing with Grafana dashboards provided by Sitespeed. Is this assumption correct?

In the case of a production Grafana setup (possibly without docker compose) I still have to check but I thought about calling `grafana-cli plugins install <plugin-id> <version>` as part of the default dashboard setup. But if I understand correctly, even if I do that, it will only affect _new_ installations of Grafana, right?

erickwilder:
I didn't have the time to look into this and our plans to use InfluxDB are on hold right now.

PR #2858, opened 2020-01-30
Add fast-crc32c back

radum:
It is used for validating the uploaded files to Google Cloud storage. This package
is not required but speeds up the validation process.

It was removed as part of ##2634 fixing Node v12 native module issues. Compatibility with Node >=12 have been fixed and this can be added back now.


soulgalore:
Thank you @radum !

Issue #2866, opened 2020-02-08
Make a climate friendly plugin


soulgalore:
I'll start by just setup a new structure for a plugin and create a branch and then iterate. I'll make the plugin run by default and then we decide how we wanna do when we release it. Either its a standalone plugin or bundled with sitespeed.io. 

Ping @mrchrisadams

soulgalore:
There's a branch now called green that we can use as a start: https://github.com/sitespeedio/sitespeed.io/tree/green/lib/plugins/green

It doesn't do anything at the moment except generates the tabs, but no data included.

mrchrisadams:
Oh @soulgalore, I'm so happy you started this!

I've been trying to work out how to build plugins for sitespeed to make this possible, but hadn't fully understood the API well enough to start myself, so this is so wonderful to see.

### Some background for using this to work out emissions on digital projects or products

I think it's worth pointing you to this spreadsheet I use as a basis for working out the emissions associated with an _entire_ digital project, including the servers and use of officespace, as it effectively shows what I've been doing manually to arrive at numbers for CO2 emissions, based on the reports and papers I've been able to read so far.

https://docs.google.com/spreadsheets/d/1dO2cFUu-IhK6wPwaemdxmWiQBn3WmD5vqnJXaA63yZw/edit#gid=2017551638

For a first version to iterate, I reckon it would be helpful to simply take figures for data transfer and convert them to carbon using an existing emissions factor, from a published report.

In more detail, I think there are three levels, where it would help to know where the carbon emissions associated with a page view are to inform decisions:

- **a page level**, to provide total co2 per page view - the spreadsheet I linked shows how I've been using sitespeed and analytics software to work out some indicative CO2 per month, or projected emissions for a year, for existing projects, to inform product decisions.
- **a domain level**, to work out where in your supply chain the emissions are occurring, so you know what to change, or which of your vendors to speak to, about not running on fossil fuels, for example. In my work at the green web foundation, we publish datasets of which providers run on green energy as a SQLite, as open data, to make this easier (infra running on green energy will have lower carbon emission than fossil fuels, which is why we track this).
- **a resource level**, so you can work out what measures to take in terms of WPO, from optimising assets, or making changes server-side to reduce what gets sent over the wire. You can see an example of this where @hanopcan helped [me make a change on a wordress theme to reduce page weight by around 90%](https://github.com/thegreenwebfoundation/tgwf2015/issues/1).

[...]

PR #2868, opened 2020-02-12
WIP - Add carbon emissions figures for data over the wire

mrchrisadams:

- [ ] Verify that the test works by running `npm test` and test linting by `npm run lint`

### Description

This PR is for #2866 - _Make a climate friendly plugin_.

This is the first pass at adding some numbers based on data transferred over the wire. It uses the 1byte model, for calculating CO2 emissions per byte of data sent, from the [Shift Project](https://theshiftproject.org/en/lean-ict-2/).

Right now, we:

- assume US grid intensity (the amount of CO2 emitted per kilowatt hour of energy)
- assume every domain is running on the same kind of power
- ignore the emissions from making the device we use to access an app/site

### Questions I have

- we can check each domain against the green web foundation API. Doing this would mean that for content form these domains we'd use a lower figure for CO2 per byte. How do I get a list of domains to add this up?
- it would be nice to let folks update the page view figures interactively, as suggested in the code. Is there an exmaple of interactivity I can look to?
- it would be good to convert to sensible figures for weight, like tonnes and kilos, instead of 1,000,000 grammes. Is there any preference on how we format numbers and so on?


soulgalore:
Hi @mrchrisadams great, I also pushed a couple of things to the branch today:
* Simple host check against the web foundation API
* Configurable number of page views with `--sustainible.pageViews 100000`  - that way users can set their own limit and then we can send that number to Graphite. I need to configure the data we want to send but I haven't done that yet, I'll do that when we feel we are more or less ready for a first release.

> How do I get a list of domains to add this up?

Checkout https://github.com/sitespeedio/sitespeed.io/blob/green/lib/plugins/sustainable/hosting.js#L43-L45

>it would be nice to let folks update the page view figures interactively, as suggested in the code. Is there an example of interactivity I can look to? 

No we don't have a perfect example today, but you can look at the tabs: https://github.com/sitespeedio/sitespeed.io/blob/e3dcd3fdb7f934d4e355bbbc52c93900cebfacf4/lib/plugins/html/templates/url/summary/tabs.pug and https://github.com/sitespeedio/sitespeed.io/blob/green/lib/plugins/html/templates/url/includes/tabScripts.js

> it would be good to convert to sensible figures for weight, like tonnes and kilos, instead of 1,000,000 grammes. Is there any preference on how we format numbers and so on?

If you can keep grammes in the code (then we send that to Graphite/InfluxDb and then you can choose format there). In the frontend in the pugs, we can add a helper method to format grammes to something more readable. Checkout the helper for bytes https://github.com/sitespeedio/sitespeed.io/blob/green/lib/support/helpers/size.js

mrchrisadams:
Awesome, thanks Peter!

I was working this morning on using the multi domain API to check _all_ the domains in a single API request, and to create a breakdown, and also provide a breakdown of CO2 by domain, based on this.

I think that's stuff we can put into a separate PR - want me to make a separate one there?

You want me to make a separate PR for this?

soulgalore:
Ah great!

We can keep it in the same, I wanna wait with merging this until we are almost ready for release, I keep master clean for now for coming patch releases. Since this is new functionality its ok to have a kind of larger PR.

soulgalore:
@mrchrisadams seems that testing a lot of domains slows down the API? Running `bin/sitespeed.js https://www.expressen.se -n 1 --headless` gives me a 500 error.

soulgalore:
@mrchrisadams I added a switch to enable the plugin, so the next time you try it add `--sustainable.enable` so it actually runs.

mrchrisadams:
Thanks Peter!

I looked up the issue about the slow API call.

For domains new to the system that aren't in the API's cache or database,
it needs to do a DNS lookup. One of the domains was taking ages to resolve,
which caused a timeout, resulting in the 500.

However, in the greencheck APi side, it did kick off an async job to look
up the domain and when it returned, cache the result.

Once this lookup was done, subsequent requests returned fast as it was
pulling from a cache.

We should have some better timeout logic on the API side, but til then, if
the multi request doesn't return inside a few seconds, I think we might be
better off throwing an exception, and letting the async job lookup the
result outside the request cycle for this run.

I'll be looking this issue this afternoon to provide the paper domain
breakdowns.




Sent from my phone. Please excuse typos and brevity. See all the ways to
contact me at https://productscience.co.uk

soulgalore:
Cool. I've added a couple things:
- to run the host check you need to add `--sustainable.hosting` that way its off by default and that kind of makes sense for privacy reasons. 
- send metrics to Graphite/InfluxDB

BTW the certificate on the API server has expired.

mrchrisadams:
Hi @soulgalore - thanks for the pointers before.

I've added a little more to this now, so we now have a datastructure for CO2 emissions by domain, passing in the lower figure for domains served from greener infra.

I've blocked out time on Friday to write a blog post on the green web Foundation website, which should have some a paragraph to re-use, but I think I need some help in a few areas.

*Code style and linting*

I'm doing something like this, to iterate through objects and filter them. The linter complains about me not using the key variable, but I'm not sure how to either tell the linter to ignore it, or the preferred style for doing this for the project.

I'm happy to adapt my code for consistency, but I didn't see a dominant pattern in the code to follow. How should I be writing this to make it fit the project better?

```javascript
const entries = Object.entries(allGreenCheckResults);

let greenEntries = entries.filter(function ([key, val]) {
  return val.green;
});

return greenEntries.map(function ([key, val]) {
  return val.url;
});
```
*Making tables in pug*

I wasn't sure of the best way how to iterate through a data structure with pug and present it as a table. I can generate the datastructure, but it wasn't obvious to make how to make this work with tables as I don't understand the median/percentile summary stuff yet.

I've outlined the rough markup I'm going for - can you let me know how I should do that?

Alternatively, if you are happy doing the templating, I'll focus on extracting the co2 calcs and hosting into separate modules to include from npm instead, and get those published as we discussed before.

soulgalore:
I will fix these tonight but just FYI:

> Code style and linting

You can setup your editor to respect the eslint/prettier setup in the project. Else you can also just run `npm run lint:fix` that fixed the lint problems (in this case it wa a missing space). I use Visual Code that autoformats on save + you can also get an error log with the errors on the file and auto fix in the editor.

> Making tables in pug

This isn't obvious: At the moment we use one pug file two display two set of values: one for each run, for example one specific metric and one for the page summary (the min/median/max) for all the runs. If you use sitespeed.io and wanna measure the performance of your site you do multiple runs and then use the median (or min or whatever) metric to remove some of the noice. At the moment we only collect some of the metrics for pageSummary so the pug file is breaking. If you have the data you can loop/make a table with 
`each val, key in sustainable.Co2PerDomain` if you have an object with keys. I'll go through add add statistics for all the metrics now.

soulgalore:
@mrchrisadams I've pushed a couple of changes + added first vs third party.

I think the numbers aren't adding up correctly in co2PerPageView but I can have a look at that tomorrow (taking first+third party co2 is smaller than  co2PerPageView  but should be the same).

Haven't done anything about co2 per resource/content type, I'm thinking "per resource" will be low so do we need it?

soulgalore:
Tonight I'm gonna polish the code, there's a couple of places where we can simplify and rename  to follow naming standards and I will also look into co2PerPageView calculation.

soulgalore:
Ok @mrchrisadams I've done a couple of changes and gonna let it be until tonight: We now calculate co2 per domain, so if we have green hosting enable, every request/domain will be checked against that list so we can green/grey per domain.

I've added the dirtiest resource "top list". Go into a specific run page (1,2 etc) and you will see that.

mrchrisadams:
Sweet! Thanks for this Peter!

I've just run a check and I think there's something up with the pug template, but I don't understand the aggregation part well enough to fix it myself.

In the pug template,  we currently have a check like so:

```
if sustainable.co2PerDomain
```

I think it needs to be like this to pick up the flags we pass in on the command line

```
if options.sustainable.co2PerDomain
```

Once we have that though, I get an error trying to look through the `sustainable.dirtiestResources` and `sustainable.co2PerDomain` collections, even when I try to follow the pattern I see elsewhere in the pug template to present the median figure.

This is something you know better than me, so I'll focus instead on getting the npm package for hosting and CO2.js, published so that can be included that way instead.

Thanks again all the help 👍

soulgalore:
Ok so if you run `bin/sitespeed.js https://www.aftonbladet.se -n 1 --headless --sustainable.enable --sustainable.hosting` that doesn't work for you?

For the template and `sustainable.co2PerDomain`: At the top you the pug file we check if we are on a pageSummary (summary of multiple runs where we have median/min ex) or a specific run and just create: ` const sustainable = pageInfo.data.sustainable.run ? pageInfo.data.sustainable.run : pageInfo.data.sustainable.pageSummary`

That way `sustainable` will have the path to the data, independent of page type BUT for pageSummary is median/min etc. We could have two different pug files, one for displaying pageSummary and one for run but if we can avoid it we don't need to have (almost) the same code on multiple places.

Per domain should by default but you need to go into a specific run:
perRun

We could jack in data from the first run to display on the page summary page (for metrics that we don't have stats) but I haven't got that far yet. I can fix that tonight.

So on the pageSummary we only show data that is collected in `aggregator.js` file at the moment. That's why you need to go into an individual run.

mrchrisadams:
Ah, thanks for clearing that up. 

Yeah, on individual page runs I now see the numbers - sweet!

It occurred to me - because we're not showing the actual data sent over the wire, next to the Co2 figures, it's not so obvious that they're coming from the page size.

In future, I think it might be worth adding this column to the table to make it easier to understand.

I'll include this in the hosting and CO2 npm module, so it's easier to access in a pug template.

soulgalore:
Yeah, I agree, I just added the transfer size so its easier to understand. I'll try to add the extra data on the first "pageSummary" page tonight + then would love a small text from you @mrchrisadams about the calculations so its clear for people how it is done.

soulgalore:
One more thing: For me we could merge this PR now and then we can do follow ups before the release and change to the npm modules? That way, I can let it run on our test instance for a while just to make sure nothing slipped through.

mrchrisadams:
Hi Peter! Yeah, I absolutely think there's enough her for a v1.

I'll make PRs for extracting the external npm modules, once they're up, and working.

I'll write a para about the calculations and assumptions being made, and where the areas for improvement from a methodology would be, and add it to the original issue.

Thanks for all your help :)

mrchrisadams:
Hi Peter, here's the paragraph - the detail below is listed in case you want to add more in the blog post, but once we have the external npm modules, I'll add another comment with a link to the blog post on the Green Web Foundation with more details.

> We know using the internet means using electricity to power servers. And because most of that electricity comes from burning fossil fuels, it means every byte sent has a cost in carbon as well as power. The sustainable web plugin combines the latest in peer reviewed science and open data from the Green Web Foundation to help you build greener, more sustainable websites and applications.

## How we do it the short, TLDR version -

We work out how much energy it takes to serve a site, then work out how much CO2 is emitted to generate the power needed that electricity, based on what information we have about where the power comes from.

For the uninitiated, TLDR stands for [too long, didn't read][tldr].

[tldr]: https://en.wiktionary.org/wiki/TLDR

## The slightly longer version*

### First, work out the energy needed

We work out how much energy it takes to serve a site, based on the 1byte model published by The Shift project in their report [Lean ICT: for a sober digital][lean-ict2].

[lean-ict2]: https://theshiftproject.org/en/lean-ict-2/

This in turn draws numbers from the paper [Projecting the chiaroscuro of the electricity use of communication and computing from 2018 to 2030][anders1]. This is the most recent peer reviewed number we could find providing in format offering a breakdown of energy use broken down by datacentre, data tansfer and device usage. When we find a more recent up-to-date model, offering a similar breakdown we will update the plugin to use it.

While the 1byte model also includes the energy usage for the production and use of devices, device we do not include this in our calculations. We do this for a number of reasons:

The first is that devices use power differently viewing a website or application compared when continuously downloading video for viewing. This is acknowledged in The Shift Project's own report.

The second is that the sitespeed plugin is intended to help people realise that performance budgets can also be carbon budgets, and they can make measurable steps to improve the climate impact of digital services, through changes in how they are made.

Third, the majority of the carbon emissions from a device come from the actual production of a smartphone or laptop, not the use. While thoughtful design choices can make it possible to run applications and websites on older hardware, it's much more difficult to measure how these changes in design affect the climate impact without collecting further information about the end user's devices. This is possible with tools like Matomo or Google analytics, but this goes beyond the scope of this plugin.


### Estimating carbon for this energy

Once we have the estimated energy usage needed to serve a site, we then need to work out the emissions from this energy one for conventional, 'grey' power, and one for renewable green power.

Larger datacentres have historically been cited close to sources of cheap power, available around the clock, which would usually come from fossil fuels previously. This often results in an [energy mix tends to be dirtier than average][beam], so we use the Shift Project's figure of 519grams of CO2 per kilowatt hour.

For the rest of the internet, we use the [International Energy Agency's][iea2018] figure from of 475g of CO2 per kilowatt hour for for 2018, as an average across the globe.

Where we know power is generated using renewables we use the figure from the UK's regulator Ofgem, for renewables including solar, wind and hydroelectrity, of 33.4 grammes of CO2 per kilowatt hour. To tell how a site is powered, we look up the domain with the Green Web Foundation's - either via the their API, or when available, using their publicly available [url2green][] dataset.

[beam]: https://the-beam.com/energy/how-weve-made-the-web-dirty/
[iea2018]: https://hyp.is/MPRiUlTMEeq13bM-WjChjQ/www.iea.org/reports/global-energy-co2-status-report-2019/emissions
[url2green]: https://www.thegreenwebfoundation.org/green-web-datasets/
[anders1]: https://www.researchgate.net/publication/331047520_Projecting_the_chiaroscuro_of_the_electricity_use_of_communication_and_computing_from_2018_to_2030

soulgalore:
Thank you @mrchrisadams this is perfect, I'll put it in the documentation for the plugin for now and then we can add a link to your post when its ready.

PR #2899, opened 2020-02-26
Extract co2 and hosting checks into separate module

mrchrisadams:

### Description

Hi Peter, this PR extracts the hosting and co2 checks into a separate npm module like we discussed before - we now import the [@tgwf/co2](https://www.npmjs.com/package/@tgwf/co2) module instead, so I've slimmed down the code in the plugin itself.

It also includes checking with SQLlite when there is a local database for faster checks. I'm happy to discuss publishing this as a separate optional module, as I appreciate hitting the API instead of a local db might be totally ok for others too.

*Minor notes on the CO2 figures*

As I've seen some more research published since we started that would affect the numbers we get from CO2 calculations I'm likely to push an update as a version bump, but I'll include that as a separate PR.

Also, if you point me to where in the documentation I ought to add a PR, I'm happy to make one there too.


soulgalore:
Great @mrchrisadams ! I’ll push a branch with the updated docs in a couple of hours when I’m back home.

Just wanted to be 100% sure: the database is included in the PR in the module, so this will not generate requests to the API?

soulgalore:
@mrchrisadams the branch for the docs (and te file for the plugin): https://github.com/sitespeedio/sitespeed.io/blob/docs-12.1/docs/documentation/sitespeed.io/sustainable/index.md

I plan to release it either tomorrow (Thursday) or Friday morning, depending on when I can finish the docs and I get the new polished logo.

soulgalore:
@mrchrisadams package.json and the lock file npm-shrinkwrap.json is not insync, if you can just pushed the npm-shrinkwrap.json file and the test can pass and I can merge, else I can fix it tonight.

mrchrisadams:
Hi Peter!
Hi Peter! 

Responses are inline (was in a few meetings, soz)

> Just wanted to be 100% sure: the database is included in the PR in the module, so this will not generate requests to the API?

By default this will look for a local copy of the sqlite database to look for. so better sqlite has the drivers, but not the database. I think the database is not part of the npm package, as I figured it might need to be updated more regularly than the code.

If it can't find the database, it will send a request to the API instead to check the status of the url(s) passed in.

I realise this means we include `better-sqlite` when we might not need it. I'd like to extract _that_ so it can be included separately … as soon as I know a sensible way to do so.

So, probably next week or in March.

I'll push the corrected `npm-shrinkwrap.json` now.

👍

soulgalore:
I see. I think it would be great if you could supply a path to the database in that case, or what's the best way to place it/push it?

From my part having the database in this package or another npm package (that's auto published when you push a new version) would be great, because then the new version number would be noticed, so we could easily update by only changing the version number in the package number. Else we (I) need to download a new version every time of the release manually and I think I will forget.

But this will work for now, as long as I know where to place it :)

mrchrisadams:
Hey Peter.

Right now, I follow the example you shared with me to do this, keeping the path in `package.json`:

```js
function getDatabase(databaseName) {

  if (databaseName) {
    log(`looking for db at ${databaseName}`)
    return new Database(`./data/${databaseName}`, { readonly: true, fileMustExist: true })
  }
  else {
    // keep the name in package.json so its easy just to change
    const DATABASE_NAME = require('../package').database_name;
    log(`looking for db at ${DATABASE_NAME}`)
    return new Database(`./data/${DATABASE_NAME}`, { readonly: true, fileMustExist: true })
  }
}
```

The path in package.json for that module is `url2green.db`:

```
"database_name": "../url2green.db"
```

I _think_ we might be able to override this in the package.json file for sitespeed itself, but I'm not too experienced with package.json, so I may well be wrong.

*A way to get the database locally.*

I hadn't thought about using npm for publishing this as a separate module, but I think we could use an [npm post-install hook][post-install] on an a hypothetical module called `@tgwf/hosting-db`, to fetch the database on install, and move the data accessing the code there.

That would mean we can keep the original co2 module small, and move the db accessing logic into a separate module. in a future PR.



[post-install]: https://docs.npmjs.com/misc/scripts#examples

mrchrisadams:
Okay, i've pushed the shirnkwrap and made a slight change for listing the domains.

soulgalore:
hmm seems like all dependencies for the sqlite package doesn't build on a couple of OS (missing make etc), let me have check tomorrow, maybe there's a better way to do it. Maybe its better to use another SQLLite module that more easily work.

For the module, I've not tested (only looked at the code) and I think the db version doesn't work, in some cases it takes the db name but some functions take the actual db. If it's ok for you I can test and do a PR with some fixes?

mrchrisadams:
Hey Peter, it looks like this module is passing on travis now, and the logs in this PR seem to point to failures in other npm modules.

https://travis-ci.org/thegreenwebfoundation/co2.js

I'm gonna ask for some pointers for breaking this into separately installable npm modules to remove the sqlite3 dependency, when it's not used, but it's not a thing i know how to do.

Is there anything else you'd want from me here?

soulgalore:
@mrchrisadams yeah but here it fails and the problems comes from dependencies of the sqlite library, if the OS where you run install doesn't have pre compiled binaries, it build it on that OS (needing make) and thats failing in our Docker container. So that's a potential thing for people installing sitespeed.io so its a little bit worrying. That's why I was thinking about looking into another more light weight lib, but I haven't had time yet.

Tonight (or actually in a couple of minutes) I got a couple of hours for hacking, I wanna do a couple PRs in your lib. I'll make some issues and wait on you comment them, and if you feel ok, I'll do PRs ok? And then tonight I hope to get the final new *secret* version of the logo and I will prepare the docs I can release tomorrow morning or whenever I get the logo :)

mrchrisadams:
Ok, sounds good. 

Thanks Peter.

FYI - I'm happy for this to go live without needing to use the hosting checks that split between green and grey. We can add that later if need be.

mrchrisadams:
okay, I've bumped the version and published it.. thanks for your help peter  :)

soulgalore:
Thank you @mrchrisadams I just pushed so we use a local db by default but also made it configurable to use the API directly through a switch.

I'm gonna merge when the test pass and then as a last fix check if I can make the dependencies optional.

One thing @mrchrisadams : Today we have the example of 1000 page views:
Screenshot 2020-02-29 at 08 34 42

I'm thinking about just remove that? We have per page view + you can configure your own amount of page views. Or should we list more examples, 100 000/million etc? What do you think?

PR #2909, opened 2020-03-05
Update the listing page

mrchrisadams:
I initially opened this because when I ran sitespeed, I was getting zero percent for some percentages - you can see them at the static list below for convenience:

https://sitespeed.thegreenwebfoundation.org/thegreenwebfoundation.org/2020-03-04-18-40-09/pages/thegreenwebfoundation.org/index.html#sustainable

I realised there were few other things I wanted to work on once in the pug file though, so this PR for the sustainable web plugin also

- adds data transfer numbers next to CO2 in the summary, like with the breakdowns, to make it easier to see where the CO2 figures came from
- revises the copy to be a bit more approachable, and show how stuff devs do already can help
- adds a call to `toLocaleString()`, to make the user provided page numbers easier to read, i.e. _1,000,000 page views_ instead of _1000000 page views_

I'm sharing this early, as I think I need a few pointers about why I'm getting a zero number for the percentages.

Peter, are you seeing this as well?


soulgalore:
Anything else you want on this one @mrchrisadams except for letting it pass Travis? I  can have look if you want.

mrchrisadams:
Hi Peter! Are you getting the zero percent figures when you run this?

I was when I opened this PR a few days back, but I haven't checked since the new merge.

soulgalore:
I haven't tried it, I only saw Travis failing. One lint issue and one variable not working. Sorry I haven't looked, I missed that :) Let me check.

lib/plugins/sustainable/index.js
```diff
@@ -150,6 +157,7 @@ module.exports = {
                 totalCO2,
                 hostingInfo,
                 co2PerDomain,
+                totalTransfer,
```
soulgalore: This isn't defined.

soulgalore:
Just want to check @mrchrisadams that you pushed everything? There's a variable that isn't define that breaks it for me.

mrchrisadams:
Hi @soulgalore. I'm looking into this now, and I'll get this passing on travis before messaging.

soulgalore:
@mrchrisadams did push now, it was missing calculating the totalTransferSize across all runs. Can you check if everything looks ok for you?

mrchrisadams:
Ah, I've juuust realised why the figures for percentages were so low. When we pass in a figure on the command line like `--sustainable.pageViews 1000000`, we kept getting figures like 0.00% in the tables:

Screenshot 2020-03-12 at 22 04 48

This was driving me crazy, figuring out why the numbers seemed off - until I realised that total CO2 is multiplied by this `pageviews` figure.

So, with one page view we end up with something like the CO2 per domain divided by the total page CO2 for a page, to give a percentage figure.

With 1000x page views, we end up with a muuuch smaller percentage, as we're dividing the CO2 per domain by the the total CO2 for the entire page _times 1000_.

I'll raise a separate issue for this, the rest of the copy changes should still be an improvement over what we have.

soulgalore:
Lets merge this and I'll fix that % in a follow up PR in a couple of minutes.

PR #2912, opened 2020-03-09
Use coach-core instead of webcoach directly

soulgalore:


drptbl:
Hey @soulgalore. After this change I'm not able to use sitespeed.io on node 10, also I can't find any information about dropping support for node 10 in sitespeed.io changelog. Is there any chance to fix it? Thanks :pray:.

```
 error coach-core@5.0.0: The engine "node" is incompatible with this module. Expected version ">=12.16.1". Got "10.19.0"
 error Found incompatible module.
```

PR #2914, opened 2020-03-10
Slim version of Docker using Firefox.

soulgalore:
At the moment this is not using xvfb, ffmpeg, imagemagick, python,
visual metrics etc. Also the sustainable plugin is not working.

At the moment this land on 739MB but it need the latest changes that fixes the broken npm-shrinkwrap.

`docker build -t sitespeedio/sitespeed.io:slim -f Dockerfile-slim .`


mrchrisadams:
Wow @soulgalore that was quick!

Can this run headless already?

soulgalore:
Yeah, I need it for work :)

Yep, it's been setup for headless. But for your use case it needs some fixes to get the sustainable plugin to work.

PR #2919, opened 2020-03-13
Calculate percentage per page view instead of total

soulgalore:
fyi @mrchrisadams  :)


mrchrisadams:
Hahahaha! 🤦‍♂️ Best PR ever :)

Thanks Peter!

PR #2922, opened 2020-03-15
Check against green domain list in JSON, instead of SQLite database

mrchrisadams:

### Description

This PR removes the need to check against a sqlite database, by updating the external co2 module dependency and replaces the older url2green database with a list of domains serialised as JSON instead.

This should mean we have no dependency on sqlite.

I'm a little unsure about having the JSON file included in the codebase as it's basically data that will change every month or so, so I'm happy to discuss fetching it from a stable, remote url instead as part of the install process.


mrchrisadams:
This seems to fail on windows. I'm not sure why, as the traceback doesn't make me think it's related to any of the sustainable web related stuff.

soulgalore:
Yeah the Windows tests are very unstable, I'll make the tests pass even if Windows fails. I'll test and merge tonight, thank you @mrchrisadams . I'll make a new sitespeed.io release ASAP when this is merged.

soulgalore:
Thank you @mrchrisadams !! This is super useful for me :)

Issue #2952, opened 2020-04-07
pageXray showing transfer and content size as same. 


kmaqsudi:
Pagexray showing both content and transfer size as same on version 12.4.0.

![image](https://user-images.githubusercontent.com/1022410/78708519-4393f980-78e0-11ea-92b1-8d0b18560d88.png)

soulgalore:
Hi @kmaqsudi what browser are you using? And what URL did you test? PageXray takes data from the HAR file, so if something is wrongly generated there, it will also be wrong here. Long time ago Firefox didn't generate correct numbers but I think that has been fixed.

Had a quick look for Chrome here: https://s3.amazonaws.com/results.sitespeed.io/www.google.com/2020-04-07-16-11-18/pages/en.wikipedia.org/wiki/Barack_Obama/index.html#pagexray

And Firefox: https://s3.amazonaws.com/results.sitespeed.io/www.google.com/2020-04-07-18-07-36/pages/en.wikipedia.org/wiki/Barack_Obama/index.html#pagexray

That looks ok for me. Can you help me reproduce your issue?

kmaqsudi:
I'm using chrome only.  I've narrowed it down to it only happening on the first URL tested when testing multiple pages.  The rest of them seem to be fine.  

![image](https://user-images.githubusercontent.com/1022410/78722544-b492db80-78f7-11ea-9a51-63f53d815789.png)

soulgalore:
Ooops I pushed that bug button a couple of times, I wanted to verify that its really a bug first.

Ok, I think the problem is related to that you test WebPageTest and Browsertime at the same time (and could be one for desktop and one for mobile), I think that's a use case I haven't thought about. At the moment PageXray handles data for both WebPageTest and Browsertime, that works fine if you run one of the tools at a time but using them together seems to be an issue.

One fix could be that if you use both in the same run, we only use data from browsertime (that's how we usually do it). I'm gonna do a new release later today with new versions of Chrome and Firefox  later today and I want that release to only go out with the new browser, but then the coming days I can do a fix for this issue, I could probably prepare a PR later today.

soulgalore:
This has been fixed.

PR #2954, opened 2020-04-08
Only collect browsertime har when you run browsertime #2952

soulgalore:


soulgalore:
Looks good on Travis, but somehow Travis can't report it to Github.

PR #2971, opened 2020-04-27
fix(plugins/slack): correctly send messages on html.finished message

### LINKED ISSUE #2147: Slack messages not being received when resultBaseURL is set
Hey guys,

strange issue for you!

When I run this:

`docker run --shm-size=1g --rm -v "$(pwd)":/sitespeed.io sitespeedio/sitespeed.io:7.3.6 https://sitespeed.io -n 1 http://sitespeed.commercegurus.com/speedreports --slack.hookUrl https://snip --slack.type summary`

I get a Slack message (webook address snipped above)

When I add a `resultBaseURL` as follows:

`docker run --shm-size=1g --rm -v "$(pwd)":/sitespeed.io sitespeedio/sitespeed.io:7.3.6 https://sitespeed.io -n 1 --resultBaseURL http://sitespeed.commercegurus.com/speedreports --slack.hookUrl https://snip --slack.type summary`

I get no Slack message. The url is valid as the report is visible at the the correct url.

Any ideas?

Running via docker on Ubuntu 16.04

Thanks

erickwilder:

### Description

Do not rely on `resultBaseUrl` to block messages to be sent to Slack. Instead, check if any
of the supported hosting providers are correctly configured. If that is the case, then
messages are sent _only after the upload is complete_. For all other cases, send the message
anyway.

**Add basic test scenarios**:  

- Check for missing options when calling `slack.open()`
- Ensure that methods supposed to return lists always return them
- Add some sanity checks to the summary, so that missing values would not crash

**Notes**:

Most test cases were included because I had to wrap my head around the whole Sitespeed architecture and how the `postMessage()` calls from the queue would end up in the Slack plugin. While doing it so, I noticed that when calling the plugin with some specific test data (i.e, missing arguments, or `undefined`) the plugin would crash. I do expect these cases to not happen when running sitespeed but I took the approach of _better safe than sorry_ here. (i.e 
 a `getURLs()` method returning `undefined` doesn't look correct as I would expect it to _always_ return a list, even if it's an empty one). 

---

Closes #2147


soulgalore:
Great @erickwilder I'll have a look tomorrow!

soulgalore:
The tests passed but Travis having problem reporting it to Github. 

Extra :gift_heart: to  @erickwilder  for adding tests :)

Issue #1065, opened 2016-07-21
Let plugins post register help parameters


soulgalore:
With the current setup, the parameters displayed with <code>--help</code> is inside of the cli. It would be cool if each plugin could register parameters for the cli, so that if you add your own plugin, the help for that would be displayed if you run: <code>sitespeed.io --help --plugins.load myownplugin</code>

erickwilder:
I'm interested in working on this one and I might spend some time trying out some ideas but before I go down the rabbit hole. Do you guys have any ideas how to go on with this one?

soulgalore:
This is possible now and implemented in the influxdb plugin.

PR #2984, opened 2020-05-06
Refactor grafana cli options

erickwilder:

### Description
This is a proposition to start moving the responsibility of plugins defining their own CLI options. In an ideal situation _all_ plugins, not only built-in ones could benefit of such structure.
This is is related to #1065 - it doesn't address that issue entirely and I decided to put this to start a conversation about how process in next steps.

#### Suggestions

- Migrate all other plugins to match the proposed format (in separate PRs)
- Implement in the CLI some logic to load options from plugins added via command line/configuration (i.e `--plugins.add`)
- Update the plugin documentation to officially support `options()` method, _after_ we have a solid strategy to include 3rd party plugin CLI options


soulgalore:
Hi @erickwilder ah cool.

I think the key issue (that stopped me to implement this before) is to find a good pattern your bullet number two:
>Implement in the CLI some logic to load options from plugins added via command line/configuration (i.e --plugins.add)

It would be good to iterate how that would work before we start move over to the new pattern.

erickwilder:
I understand. I wanted to first work in the loading logic but as soon as I started digging in the cli code I saw that the codebase in there had a lot of distracting parts that could easily be moved over to the plug-in. The rationale here would be of cleaning up the code for cli _first_, then one could better focus on getting external plugins to define options.

Adding these options to the built-in plugins already has the benefit of a better separation of concerns. 

As for the loading strategy, I thought of separating the CLI processing in "phases" or "tiers" 

- Top-level options `--plugins.*` would be evaluated _before_ any other option
- Plugins listed in the configuration would probably have to be loaded in a similar way the `loadPlugins()` function work, but yielding just the plug-in module
- With a list of loaded plugins we could apply the concepts introduced in this PR to each one of them

I would ask you to consider moving the options to built-in plugins even without any dynamic loading of plugins. This is already the case for plugin authors: they have no way to add help messages in the standard Sitespeed help messages. Even if we have a new (for now undocumented) method, there is no regression but we pave the way to make that happen

soulgalore:
Yep I see your point, sounds good. I also like the idea that each plugin handles the cli options.

lib/plugins/grafana/index.js
```diff
@@ -1,11 +1,57 @@
...
+
 const defaultConfig = {
   port: 80
 };
+
 module.exports = {
+  name() {
+    return path.basename(__dirname);
+  },
+  options() {
```
soulgalore, line 15: What about move those options out to a new file (cli.js or something) and set a new standard for that also? For some plugins the index.js is cluttered I'm thinking about minimising the code in there?

erickwilder: I updated with the following:

* Renamed the proposed method to `cliOptions()` to be more explicit and to not shadow/override the property defined in the `.open()` method.
* Moved all definitions to `grafana/cli.js` that's later on `require`d by the method/property above

Not all plugins will have extensive options (like `browsertime`) and I think it's very convenient to keep the plugin interface clear with an explicit method/property like `cliOptions()`

erickwilder:
@peter please, check my [latest commit](https://github.com/sitespeedio/sitespeed.io/pull/2984/commits/048cc71365d8595f666890161c9923fb7c104f5a). I took your suggestion into account _and_ decided to spend a bit more time to get some things prepared for upcoming PRs touching other plugins:

- Add some utility methods to `cli` so that it's easier to migrate built-in plugins to this format
- Add more tests to ensure nothing in the plugin interface (grafana only) related to these changes were broken
- Make `plugin.config` based on the yargs configuration (for all options defining an explicit `default:` property) instead of hard-coding them)

Let me know if this is good to go. Thanks!

soulgalore:
Wow looks great @erickwilder !

Issue #2976, opened 2020-04-30
URL alias broken in 12.7.1


kmaqsudi:
I have single URLs with an Alias (no groups) and the page summary no longer shows the Alias.  In addition, my graphite graphs are broken because the $Page alias is the full URL instead of the Alias.

soulgalore:
Hi @kmaqsudi let me try to fix that later today, there was bugs introduced in 12.7.0 but it seems its still in 12.7.1 then.

soulgalore:
@kmaqsudi I've just released 12.8.0 where there's a fix, please try it out when you have time.

edevlin-made:
Hi @soulgalore , thanks for working on this brilliant project.  I'm using the latest release 12.10.0 and still see a problem with URL aliases.  My URLs and their aliases are provided through a text file.  

I see the aliases in the "page" tag on the metrics going to InfluxDB (great), but they aren't used on the pages.html page of the HTML report - instead I see the full URL in the first column of the table.  My URLs are long with ugly querystrings, so I'd much prefer to see the alias instead.

Not sure if this is intended behaviour - feels like it regressed in 12.7.0 and is only partially fixed?

soulgalore:
Hi @edevlin-made thanks :) Yes that looks like a regression, let me try to fix it tonight.

soulgalore:
@edevlin-made I just released 12.10.1 with a fix for it. Thank you again for reporting!

edevlin-made:
Brilliant thanks @soulgalore , v12.10.1 works for me, the report looks much better now.

PR #2994, opened 2020-05-14
Fix showing alias in the HTML

soulgalore:
https://github.com/sitespeedio/sitespeed.io/issues/2976#issuecomment-628568308


soulgalore:
The tests passed but was not reported back to GitHub.

PR #3005, opened 2020-05-20
feat(plugins): move influxdb CLI options to the plugin defintion

erickwilder:
Move `influxdb` plugin CLI options into the plugin defintion/structure.

---

It follows the same pattern implemented for both Graphite/Grafana


soulgalore:
Thank you @erickwilder

PR #3010, opened 2020-05-22
Fix config sending per run

soulgalore:


soulgalore:
Hmm checking the old version and think _graphite.experimental.perIteration_ never worked as intended.

soulgalore:
Let me fix this for the next major release.

PR #3025, opened 2020-06-02
Add information about android power usage testing.

gmierz:
This patch adds some documentation on the new support for android power testing. It also outlines the metrics that could be retrieved with this, and explains how someone might disable/enable charging on their phone.


soulgalore:
Thank you @gmierz !

PR #3085, opened 2020-07-10
Send the performance budget message in the queue for easier use

soulgalore:


soulgalore:
Sometimes the action doesn't pickup that's there a new Edge.

PR #3087, opened 2020-07-15
Be less strict about GCS options in respect of ADC

scheying:
Within Google Cloud, services can make use of Credentials-"Auto Discovery" (ADC). The @google-cloud/storage library supports this and tries an automatic discovery when keyFilename is unset. In an auto-discover scenario projectId is discovered as well so this should also be allowed to not be set. The --gcs.bucketname option could be kept as a (mandatory) signal to use GCS at all.


soulgalore:
Thank you @scheying I'll merge this before the next release in a couple of days.

soulgalore:
FYI: The error in the action is because somehow our CruX secret code gets invalidated after a while, I'm gonna look into that after I'm back from vacation in a month.

scheying:
Hi @soulgalore, great - thanks a lot and enjoy your vacation!

Issue #3099, opened 2020-07-24
No unit for the slack summary


lbilhaut:
This is not a bug, but as a physicist at heart, it makes me sad that the Slack report doesn't have any unit in it. I've made the change to the file on my personal repo, [here is the commit](https://github.com/lbilhaut/sitespeed.io/commit/c3fc35d1b492ecd3a687ea5c83ce329419d23161). Is it ok if I open a PR with it? (I've never made a contribution to an open source project before so I'm walking on eggshells here!)

soulgalore:
Hi @lbilhaut looks good! After we added support for Slack we also introduced a way to format metrics in ms see https://github.com/sitespeedio/sitespeed.io/blob/main/lib/support/helpers/time.js#L36-L43 I think we also could use that there, so each metric just have a format function?

But you could do a PR first with what you done (just move the space from " ms" and have that once where you combine the metric with the unit. 

In the long run I wanna start to use https://github.com/sitespeedio/sitespeed.io/blob/main/lib/support/friendlynames.js in Slack (where each metric and formatter is already defined). I'm gonna do a rewrite of the Slack plugin when I add Matrix support in a couple of weeks.

lbilhaut:
@soulgalore  Great, thanks! The latest option sounds definitively the best path!
I've pushed my changes but I'm still having an issue with the build for Linux browsers. This is not my OS so I'm not familiar with it. I'm wondering if it's because the API key is not valid for CruX (in that case, how to fix it?) or if it's because now, it expects a space between the number and the unit.

soulgalore:
This is fixed.

PR #3102, opened 2020-07-24
Add unit to slack summary

lbilhaut:
### Description

This PR adds units to the Slack summary (issue https://github.com/sitespeedio/sitespeed.io/issues/3099)

### DIFF
--- a/docs/documentation/sitespeed.io/browsers/index.md
+++ b/docs/documentation/sitespeed.io/browsers/index.md
@@ -104,6 +104,21 @@ Chrome has a [long list](https://peter.sh/experiments/chromium-command-line-swit
 
 When you add your command line switched you should skip the minus. For example: You want to use ```--deterministic-fetch``` then add it like ```--chrome.args deterministic-fetch```.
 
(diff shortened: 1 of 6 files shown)


lbilhaut:
Great, thanks! I've pushed the changes but I'm still having an issue with the build for Linux browsers. This is not my OS so I'm not familiar with it. I'm wondering if it's because the API key is not valid for CruX (in that case, how to fix it?) or if it's because now, it expects a space between the number and the unit.

soulgalore:
Hi @lbilhaut thanks! Can you just exclude the shrink-wrap from the commit and I'll merge. 

The test failing is the Crome User Experience key that sometimes just fails, so its not related to your change.

soulgalore:
Thank you @lbilhaut !

lbilhaut:
Great, thanks @soulgalore ! That's my very first contribution to an open source project!!! (small step for sitespeedio but a big leap for me!)

soulgalore:
Great :D It will take some time until I release it though, probably next week so I can bundle the release with some more changes. Thanks again!

PR #3109, opened 2020-07-28
Update Docker usage in docs for compatibility w/ PowerShell

fcFn:
### Description
Moving the quote to the end of the bind mount command makes it work with both Bash and PowerShell.

I don't think such a trivial change requires opening an issue. Please correct me if I'm wrong.

This could also be changed in the website quick start guide for docker. Please point me in the right direction to submit the fix.

```powershell
PS docker run --rm -v "$(pwd)":/sitespeed.io sitespeedio/sitespeed.io:14.1.0 https://www.sitespeed.io/
docker: invalid reference format.
See 'docker run --help'.
PS docker run --rm -v "$(pwd):/sitespeed.io" sitespeedio/sitespeed.io:14.1.0 https://www.sitespeed.io/
Google Chrome 84.0.4147.89
{...}
```


fcFn:
Ugh, I shouldn't be rushing like that. It doesn't actually work the same in bash.

fcFn:
OK, **now** it does work with both PowerShell and Bash.

soulgalore:
Thank you @fcFn !!!

PR #3139, opened 2020-09-01
#3138 - fix set aliases in cli & refactoring

### LINKED ISSUE #3138: Error  run with scripts and set aliases in cli
command for example
```docker run --rm -v "$(pwd)":/sitespeed.io sitespeedio/sitespeed.io -n 1 --verbose --multi --config app/config.json "scripts/beruPdp.js"```

Code in **config.json**
```
{
  "browsertime": {
    "cpu": true,
    "viewPort": "1920x1080",
    "iterations": 1
  },
  "urlAlias": "beru_pdp",
  "plugins": {
    "remove": [
      "html"
    ]
  }
}
```

code in **beruPdp.js**

```
module.exports = async function(context, commands) {
  commands.meta.setTitle();
  const driver = context.selenium.driver;
  try {
      await commands.measure.start();
      await commands.navigate('https://beru.ru/product/smartfon-apple-iphone-11-pro-256gb-serebristyi-mwc82ru-a/100773394793');
      await commands.measure.stop();
      const isCapthca = await commands.js.run('return (document.location.pathname.includes("showcaptcha"))');
      await commands.measure.add('hasCapthca', isCapthca);
  } catch (e) {
      commands.error(e)
  }
};
```

This works without `urlAlias`

### PR DESCRIPTION
**Pull request by pan-alexey**
### Description
fix #3138 

Add check urls if js file;
Refactoring (Remove closure in sitespeed.js)

### DIFF
--- a/lib/cli/util.js
+++ b/lib/cli/util.js
@@ -40,6 +40,12 @@ module.exports = {
       } else {
         const filePath = path.resolve(url);
         try {
+          // Check extention file is js
+          if (path.extname(filePath) === '.js') {
+            allUrls.push(url);
+            continue;
+          }
+
           const lines = fs
             .readFileSync(filePath)
             .toString()
--- a/bin/sitespeed.js
+++ b/bin/sitespeed.js
@@ -8,6 +8,7 @@ const cli = require('../lib/cli/cli');
 const sitespeed = require('../lib/sitespeed');
 
 async function run(options) {
+  let budgetFailing = false;
   process.exitCode = 1;
   try {
     const result = await sitespeed.run(options);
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by soulgalore** [comment]
Hi @pan-alexey thanks, let me test it out the coming days!

**Comment by pan-alexey** [comment]
@soulgalore nice and thank you!

File: lib/cli/util.js [hunk 1]
```diff
@@ -40,6 +40,12 @@ module.exports = {
       } else {
         const filePath = path.resolve(url);
         try {
+          // Check extention file is js
+          if (path.extname(filePath) === '.js') {
```
soulgalore, line 44: Think this will break for people that do not use extension `.js.` At the moment I think we can have whatever extension?

soulgalore:
Nothing happened here so closing.

PR #3157, opened 2020-10-07
Run tests also on NodeJS 14

soulgalore:


soulgalore:
Merging since its unrelated errors.

PR #3163, opened 2020-10-20
Fix broken link to screenshots in meta data and Slack.

soulgalore:
As reported in https://github.com/sitespeedio/compare/issues/108


soulgalore:
The test that not pass is unrelated to the change, lets fix that in another PR.

PR #3258, opened 2021-01-18
Support upload latest file for GCS.

soulgalore:


soulgalore:
Hi @beenanner added so we upload latest files if it exists. I've verified that it works except the cache settings, I've run into problem that the account need to have delete access:
`XXX does not have storage.objects.delete access to YYY`

soulgalore:
I verified that it works now with your change!

Issue #3241, opened 2021-01-09
--chrome.blockDomainsExcept not working with REPLAY=true 


SunnyGurnani:
**Without replay 3rd Party Domains get blocked as expected**
`docker run --cap-add=NET_ADMIN --rm -v "$(pwd):/sitespeed.io" sitespeedio/sitespeed.io:16.2.1 -n 1 -b chrome --chrome.blockDomainsExcept www.realtor.com https://www.realtor.com
`
![image](https://user-images.githubusercontent.com/943576/104077513-deeb2f80-51ce-11eb-9eb6-5682affed3cf.png)

**With Replay=true domains are not blocked**
`docker run --cap-add=NET_ADMIN --rm -v "$(pwd):/sitespeed.io" -e REPLAY=true -e LATENCY=100 sitespeedio/sitespeed.io:16.2.1 -n 1 -b chrome --chrome.blockDomainsExcept www.realtor.com https://www.realtor.com
`
![image](https://user-images.githubusercontent.com/943576/104077475-bbc08000-51ce-11eb-870f-cb5f91c52d5e.png)

soulgalore:
I think the setup where we route the traffic to WebPageReplay from Chrome collide when you try to block third parties, so to make it work we need to change how we do the blocking in this case.

inderpartap:
@soulgalore If we update the [start.sh-L81](https://github.com/sitespeedio/sitespeed.io/blob/b9703742849a684b5e3a8d1fea1647d2879da455/docker/scripts/start.sh#L81) to have the EXCLUDE set to the URL we are testing (basically blocking 3rd party), it just throws an error during REPLAY `ERROR: Could not load URL https://www.realtor.com UrlLoadError: Could not load https://www.realtor.com is the web page down?`

What could be going wrong here?

**Command:** `docker run --cap-add=NET_ADMIN --rm -v "$(pwd):/sitespeed.io" -e REPLAY=true -e LATENCY=100 perf-sitespeed.io -n 1 -b chrome --chrome.blockDomainsExcept www.realtor.com https://www.realtor.com`

**start.sh L81:** 
[...]

PR #3329, opened 2021-03-29
Fix replay issue with third party domains

inderpartap:
Issue - #3241 

- [ ] Update the documentation https://github.com/sitespeedio/sitespeed.io/tree/main/docs in another PR

Got the following errors with npm test (not sure if they are related)-
1) should report error
2) should contain grafana options
3) should contain graphite options
4) should contain influxdb options

### Description
Fixing bug where `--chrome.blockDomainsExcept` flag collides with the WebPageReplay functionality.

`blockDomainsExcept` uses the same functionality which is used to route the traffic back to the WebPageReplay server. These overlap and collide, which results in 3rd Party domains still coming in when `REPLAY=true`.

#### Fix:
We would dynamically generate the `execnode` command to handle the blockDomainExcept parameters. Currently, we route all the traffic (*) to localhost during replay, which is the main reason why 3rd party domains are still working. 
The fix is to create an array of all the domains provided with the `--chrome.blockDomainsExcept` flag and map only those domains to the localhost. This would only allow traffic from first-party domains to localhost and block all others.

In the scenario where the replay is true but the blockDomainsExcept flag is not set, we would use the existing functionality to map all traffic to localhost.


soulgalore:
Hi @inderpartap cool, let me test it, I have been busy with some other stuff the last days.

soulgalore:
Looks good. To make it easier to maintain I wanna move the code to NodeJS (I think we could integrate it into Browsertime directly), I'll have a go this weekend, if it works, I'll give you the cred :) If to doesn't I'll merge this. Thanks for contribution, really cool!

soulgalore:
Hi @inderpartap inspired by your fix I fixed it directly in Browsertime in https://github.com/sitespeedio/browsertime/pull/1532 . The reason is that maintaining things in bash is harder than NodeJs (at least for me). Early versions of sitespeed.io was written in bash :)

All cred to you @inderpartap I'll thank you in the changelog for your fix!

inderpartap:
Thanks a lot, @soulgalore. I wasn't sure about how to update the `execnode` command through NodeJs 😅 but that's a nice solution. 

Glad I could help.

PR #3352, opened 2021-04-19
Update scripting documentation to cover new commands in scrolling, mouse events, window switches, and navigation.

dpalmeiro:
The PR in https://github.com/sitespeedio/browsertime/pull/1533 added several new commands that can be used in browsertime scripts.  This patch updates the documentation to cover these commands.


soulgalore:
Great, thank you @dpalmeiro !

PR #3390, opened 2021-05-24
Introduce Matrix message when test is done

soulgalore:


soulgalore:
Think we can skip this.

PR #3393, opened 2021-05-26
Run tests on NodeJS 16

soulgalore:


soulgalore:
Lets push this since its a problem with GitHub actions getting new versions out.

Issue #2758, opened 2019-11-07
keep selected tab open accros runs


matti:
currently changing the run resets the active tab:

![Screen Recording 2019-11-07 at 13 57 54](https://user-images.githubusercontent.com/3893/68387478-155e9200-0167-11ea-8ea7-e6e05f1d7d13.gif)

soulgalore:
Yeah that is annoying! Hope someone with some front end skills can help us with that!

AndresContreras96:
Hello,
I would like to work on this issue, is it still available?

matti:
yes go ahead

Aadi-27:
I would like to take this issue if no one is working on it ?

matti:
go ahead

Aadi-27:
sorry, picked other open source and got hands full. Don't mind

matti:
never minded. for all future askers, just do the damn PR.

JoseNavy:
Hello! Is this issue still happening? If so, I want to help! I'm new to Open Source and would be glad to help.

soulgalore:
Hi @JoseNavy yes that would be great, I can guide you if you need help. Either you can ask here or join the Slack channel: sitespeedio.herokuapp.com

JoseNavy:
Thank you @soulgalore! I'll install the project and If I have any questions, I'll drop them here.

amtanq:
Please take a look at my pull request for the issue.
Thank you

soulgalore:
This has been fixed.

PR #3409, opened 2021-06-16
#2758 - fix: keep selected tab open across runs

amtanq:
#2758 stated the issue of resetting the tab upon page change. This pull request adds a proxy over the navigation links and specifies new page to open with proper tab.

### DIFF
--- a/lib/plugins/html/templates/url/includes/tabScripts.js
+++ b/lib/plugins/html/templates/url/includes/tabScripts.js
@@ -1,7 +1,24 @@
 window.addEventListener('DOMContentLoaded', function() {
   let tabsRoot = document.querySelector('#tabs');
 
-  let currentTab = tabsRoot.querySelector(location.hash || 'a');
+  let navigationLinks = document.querySelectorAll('#pageNavigation a');
+  for (let i = 0; i < navigationLinks.length; ++i) {
+    navigationLinks[i].addEventListener('click', event => {
+      if (!location.hash) return;
+      event.preventDefault();
+      location.href = `${event.target.href}${location.hash}_ref`;
+    });
+  }
+
+  let currentTab;
+  if (location.hash.endsWith('_ref')) {
+    const targetHash = location.hash.slice(0, -4);
+    currentTab = tabsRoot.querySelector(targetHash);
+    setTimeout(() => history.replaceState({}, location.hash, targetHash));
+  } else {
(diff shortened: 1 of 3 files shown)


amtanq:
@soulgalore In my previous pull request, there was the issue of conversion of LF to CRLF, I fixed that issue but in current PR one test linux fails named `Run tests with CruX`. It states that the API key not valid but I don't think my PR has anything related to API key for Crux can you please take a look.
Thank You

soulgalore:
Hi @amtanq sorry about that test, that test use an API key that only seems valid for members of the organisation, I haven't had to look how to set it up so it works for all users. 

Thanks a lot for the fix, I'm gonna release it tomorrow when I had some time to test it!

soulgalore:
Hi @amtanq it looks great. 

I wonder if there's a way to make it work a little bit better  for Chrome? I usually do not use Chrome but I always test it there. If I run a test with many runs for a site with many request (you can test with  `https://www.cnet.com -n 11`) then the lag between showing the summary page and then going to the right tab is much longer than using Firefox/Safari (at least for me. Do you see the same?

amtanq:
@soulgalore I confirm the issue of laggy switches on Chrome. But the issue is arising from the bigger size of generated pages and not an issue in the code of switching tabs.

I have some findings, please verify:
- Size of generated HTML is over quite large, hence the amount of time required for the **DOMContentLoaded** event to get triggered is significant.

- The image verifies that the bottleneck of the issue is slow firing of **DOMContentLoaded** event whereas the page initialization takes well under 1ms.


Benchmark Code:
```javascript
// Started javascript execution
console.time('DOMContentLoaded')

window.addEventListener('DOMContentLoaded', function() {

  // Time taken to fire DOMContentLoaded
  console.timeEnd('DOMContentLoaded')

  // Started subroutine of page switch
  console.time('PageSwitch')
  ...
}

function selectTab(newSelection, updateUrlFragment) {
  ...

  // PageSwitch complete
  console.timeEnd('PageSwitch');

  return false;
}
```

Hence, I think this issue needs a dedicated request as it is not overlapping with the current issue. I'll be happy to get assigned for this issue.

soulgalore:
Thanks @amtanq yes agree, let keep that in a another issue. The blob of HTML is large. It's better when you serve the HTML from a server and configure the HAR to not be included in the file (`--html.fetchHARFiles true`).

PR #3454, opened 2021-09-03
Add option for crawler to ignore robots.txt

dammg:
### Description
For example we have an internal test site (a sort of showcase of all our modules), that has a noFollow rule on all its pages. With that the crawler refuses to discover any pages. However there is an option in the crawler to ignore the robots.txt. This is basically my attempt at passing that option through. I have this currently running as a patched version on our site.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -1007,6 +1007,11 @@ module.exports.parseCommandLine = function parseCommandLine() {
       describe:
         'Discard URLs not matching the provided regular expression (ex: "/some/path/", "://some\\.domain/"). Can be provided multiple times.',
       group: 'Crawler'
+    })
+    .option('crawler.ignoreRobotsTxt', {
+      describe:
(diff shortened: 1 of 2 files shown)


soulgalore:
Hi @dammg ah great, can you please add that option the cli (that way it will be visible in the help). You can find the cli options for the crawler here:  https://github.com/sitespeedio/sitespeed.io/blob/main/lib/cli/cli.js#L988-L1010

Let me know if you need any help! Thanks for the PR!

soulgalore:
There are some lint issues, best case you can easiest fix it by running `npm run lint:fix` and push the change again. Thanks!

dammg:
Hi @soulgalore , sorry I don't have access to an IDE right now, so I'm doing this via the github GUI. I added the new option to the cli and hopefully fixed the linting issues.

soulgalore:
Thank you @dammg ! This goes out in the next release, probably on Sunday!

PR #3472, opened 2021-10-07
Allow crawler to also send the configured cookies

dammg:
The crawler should open pages with the same setup in order to get full results. In my case an authentication cookie is needed, to properly open the page and see its full content (including crawlable links).


soulgalore:
Thank you @dammg for the PR!

PR #3510, opened 2021-11-19
Run tests on Firefox/Chrome beta

soulgalore:


soulgalore:
We do that in Browsertime, that should be enough.

PR #3559, opened 2022-02-07
removed "$"

Bhuvnesh875:
Removed "$" from project setup commands as it is already there in the terminals ,hence creates a mess when we directly copy pastes the commands

### Description
Not a big issue


soulgalore:
Thank you @Bhuvnesh875 !

PR #3679, opened 2022-06-23
plugins/html: Fix typo with inline CSS handling

josh-lr:
This fixes a minor typo in `HTMLBuilder` that prevents plugins from adding custom CSS to the rendered HTML output.


soulgalore:
Thank you  @josh-lr !

PR #3712, opened 2022-08-22
Changes points generation for influxdb

Damien-Ar:
### Description

Reduce the number of point created in influxdb by grouping point with the same tags

Change measurement and fields for some points, take with caution (but I think there should no longer be some weird combination like "values"/"value")


soulgalore:
Thanks @Damien-Ar I'll have a go after I tested the other PR, I haven't used Influx in a while so I need some startup time ... :)

eQza:
Any love for this? <3

soulgalore:
So I'm rewriting the tags support so it will work for Prometheus/InfluxDB and then InfluxDB users will need to use a switch to us the new setup. I think that is the best way forward.

PR #3736, opened 2022-09-28
Update to 0.10.4 co2 and make it possible change model.

soulgalore:
Hey @mrchrisadams I started to upgrade to latest released version 0.10.4 and then adding the ability to choose model (but keeping 1byte as default), I think that could be a good start.


mrchrisadams:
This looks good to me, thanks Peter!

It might be useful to make it explicit in the sustainable web tab which model is in use, but to be honest, I suspect we'll after the switch to the SWD as the default, I'm not sure how many people will want to switch to the 1byte model if they aren't already using it.

:shipit:

mrchrisadams (approved):

PR #3790, opened 2023-03-16
add test for influx v1.8 and v2.6.1

boitcex:
### Description
Add tests for verifying integration with a new 2nd version of InfluxDB

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -64,3 +64,7 @@ jobs:
       run: bin/sitespeed.js http://127.0.0.1:3001/simple/ -n 1 --graphite.host 127.0.0.1 --xvfb
     - name: Run test without a CLI
       run: xvfb-run node test/runWithoutCli.js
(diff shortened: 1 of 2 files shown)


soulgalore:
Hi @boitcex 1000*thanks for adding those!

boitcex:
@soulgalore I noticed the strange behavior for the test "Run test with Influx 2.6.1".
It is failed at the main branch with the error: "{"code":"unauthorized","message":"Unauthorized"}"
But when I run the same code at sitespeed built from "Zenika:update-influxdb" all is good

soulgalore:
@boitcex ok, so just so I get it right: do you mean the code in main maybe is broken or that when the container run in the GitHub action here it won't work?

soulgalore:
Ok, I'll have a look tonight, thank you!

soulgalore:
There was a bad merge made by me, it's fixed now and it works, thank you @boitcex !

PR #3875, opened 2023-06-09
extend crux request with proxy handling

diginote:
### Description
extend crux plugin to allow crux requests behind a proxy


soulgalore:
Hi @diginote thanks for the PR, Is there a way you can add a test in GitHub action to run through a proxy, so we can make sure it continues to work if we make changes?

diginote:
hi @soulgalore,
I have updated the crux-test.yml with a publicly available proxy server. Hope this helps and is what you needed.

PR #3888, opened 2023-06-20
Fix browserScript structure.

soulgalore:
Oh, we had a really strange behaviour where we added all the Browsertime data inside the browserScript element (the data we collect with JS) and that made it so we pushed the HAR file inside that element. This fixes that and make sure we just cherry pick the exact data we pass on.


soulgalore:
This needs to go in a new major if someone was listening to the browsertime.pageSummary events.

Issue #3892, opened 2023-06-27
Graphite Web Proxy Support (Annotations)


Shalankwa:
### Feature/improvement

Currently, there is no way (that I can see) to send annotations to a graphite web instance that is behind a proxy layer; changing the full path to the /events endpoint. Trying to do this by setting the --graphite.webHost flag to include the proxy path causes an error on 'getaddrinfo'. 

`{"errno":-3008,"code":"ENOTFOUND","syscall":"getaddrinfo","hostname":"<host>/<proxy-path>"}
    at start (file:///usr/src/app/bin/sitespeed.js:146:15)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at runNextTicks (node:internal/process/task_queues:64:3)
    at process.processImmediate (node:internal/timers:447:9)
    at async file:///usr/src/app/bin/sitespeed.js:178:1`

If there could be a way to overwrite the default PATH to the graphite [events](https://github.com/sitespeedio/sitespeed.io/blob/main/lib/plugins/graphite/send-annotation.js#L98) endpoint then graphite web instances that are behind a proxy can be better supported.

Suggestion:
adding a config flag like "--graphite.proxyPath" that would be added to the **webHost** url on requests to graphite

soulgalore:
Hi @Shalankwa thanks for creating the issue, can you give me a full example of what it will look like? I guess the path could be /events by default and then just changed if you set the parameter?

Shalankwa:
> Hi @Shalankwa thanks for creating the issue, can you give me a full example of what it will look like? I guess the path could be /events by default and then just changed if you set the parameter?

Yes I think /events as the default, and only different when the param has been set.

currently our graphite web event endpoint sits at a full path like: www.domain.com/api/v1/graphite/events/, so the params we would provide in this example are
--graphite.webHost = www.domain.com
--graphite.proxyPath = /api/v1/graphite/

Within the send-annotation.js it would add the proxyPath to the default /events/. It doesn't look like you use the webHost config param anywhere else, but this way would be best to ensure you don't need to refactor anything for the "getaddrinfo" validation that is performed on the domain.

Alternatively, you could allow adding paths in the webHost config param and parse any added path off of it before validating the domain.

Shalankwa:
I created an example PR with the changes that would allow this to work... I apologize if I missed any contributing rules.

Shalankwa:
[...]

PR #3893, opened 2023-06-27
add proxyPath config for graphite web endpoint

Shalankwa:
### Description
Add support to alter the URL path to graphite-web events endpoint when behind a proxy.

Tested with my own instance of graphite-web running behind a proxy.


Shalankwa:
Related issue:
https://github.com/sitespeedio/sitespeed.io/issues/3892

PR #3896, opened 2023-07-03
Docker compose: grammar fix

Amerousful:
It's a not big deal, but I decided to fix it.


Amerousful:
Crux check was failed. 
` bin/sitespeed.js -b chrome -n 1 --crux.key  https://en.wikipedia.org/wiki/Main_Page --plugins.remove browsertime`
Does instance have `secrets.CRUX_KEY`?

soulgalore:
Hi @Amerousful thanks, that key doesn't work for contributors, that's annoying, I wonder if there's a way to fix that.

PR #3931, opened 2023-08-07
Extra safe check for annotations.

soulgalore:


soulgalore:
Unrelated Microsoft error, lets merge.

PR #3957, opened 2023-09-01
Try/catch coach/third party tools errors

soulgalore:
The reason for this fix is that the bidhar broke with Firefox 117, but maybe this is still a good idea.


soulgalore:
I think we do not need to be this careful.

PR #4041, opened 2023-12-28
Add extra guard

soulgalore:


soulgalore:
I think the root cause here is a bug with getting sustainable data, let me fix that first.

soulgalore:
There's a better fix in https://github.com/sitespeedio/sitespeed.io/pull/4042

Issue #4174, opened 2024-06-05
InfluxDB 2.x - Send annotations


Amerousful:
### Feature/improvement

Initial topic - https://sitespeedio.slack.com/archives/C03DHTPSA/p1716991642642359
Purpose: Send annotation for InfluxDB 2.x isn't implemented. Currently, it sends only for InfluxDB 1.x

Amerousful:
https://github.com/sitespeedio/sitespeed.io/pull/4197 Merged

PR #4175, opened 2024-06-05
InfluxDB 2.x - Send annotations

Amerousful:
https://github.com/sitespeedio/sitespeed.io/issues/4174

I have concerns about realization. I don't like this fact that there's two almost the same code in `lib/plugins/influxdb/send-annotation.js` and `lib/plugins/influxdb/send-annotationV2.js`, but I did it by analogy with `lib/plugins/influxdb/sender.js` and `lib/plugins/influxdb/senderV2.js`. I followed the source structure.


soulgalore:
Hi @Amerousful sorry for being late on this. So there's an error in the Linux browsers GitHub action can you check what's wrong? I can help out in the end of this week so it can go out, as long as it pass the tests.

soulgalore:
For the duplication of the code, I think its ok for now so we at least have something that works.

Amerousful:
@soulgalore Hi, don't worry.
Wow, indeed. I overlooked it. Thanks!) 
Now, it fixed.

Got it. Do not hesitate to ask me refactoring it, in the future.

PR #4197, opened 2024-06-21
InfluxDB 2.x, Annotations. Send tags for filtering

Amerousful:
Initial thread in the Slack - [link](https://sitespeedio.slack.com/archives/C03DHTPSA/p1718375497015779)

**Old**: `events title="Sitespeed.io",text="${message}",tags=${influxDBTags} ${timestamp}`
**New**:`annotations,${influxDBTags} title="Sitespeed.io",text="${message}",tags=${grafanaTags} ${timestamp}`
***

Fixes:
1) Send annotations **also** as a `tags` (not **only** as `field`'s value)
2) Rename `events` to `annotations` for appropriate naming.

~~1) Send annotations as a `tags`, because the current logic sends them as `fields`.~~
~~3) Represent tags with keys, in format `key1=value1,key2=value2`, instead of `value1,value2`.~~
~~3) Rename `events` to `annotations` for appropriate naming. However, it might not be a good idea to change it because it would be inconsistent with the underlying logic in others "senders".~~
image 3626


soulgalore:
Hi @Amerousful nice thank you. Can you test that you can get the annotations and match them like this:
Screenshot 2024-06-23 at 18 54 33

Amerousful:
Now it works. I realized that it turns out that the original tags were for Grafana. I returned it back, and now this logic writes both of them: for Grafana and for InfluxDB.
***
image 3660
image 3661

image 3659

PR #4338, opened 2024-12-16
Try to catch different number of elements when using Wilcoxon

soulgalore:


soulgalore:
Lets make a check instead if we are using wilcoxon.

PR #4341, opened 2024-12-18
[Compare] Add validation for dataset lengths for the Wilcoxon test.

Amerousful:
Add validation for prevent an error from the Wilcoxon test when dataset has different number of values.


soulgalore:
Thank you @Amerousful

PR #4357, opened 2024-12-27
Remove webdriver manager.

soulgalore:
We don't use webdriver manager and its automatically installed for Linux, Windows and Mac. Remove it.


soulgalore:
Hmm the manager is needed on Mac (else you get an error even though the driver is bundled in the OS). Lets remove the ones we don't need in the Docker container at least.

Issue #4453, opened 2025-02-27
Sitespeed and standard plugins crashes when unable to access website url


7h3Rabbit:
### Have you read the documentation?

- [x] Yes, I've read the [how to make a reproducable bug guide](https://www.sitespeed.io/documentation/sitespeed.io/bug-report/)
- [x] Yes, I've read the [how to debug my script guide](https://www.sitespeed.io/documentation/sitespeed.io/scripting/#debug)

### URL

https://webperf.se

### What are you trying to accomplish

We use SiteSpeed in our webperf_core repo where we, among many other things, try to check what TLS versions a website support to give good feedback on it it should be enabled/disabled.

Currently, when a website is not supporting the the TLS version we force use of, Sitespeed does the following:
- It retries connecting to the website multiple times (even when specifying -n 1)
- It crashes in standard plugins and internal Sitespeed code
- It bloats the terminal with a LOT of stacktraces

### What browser did you use?

Firefox

### How to reproduce

```shell
We run the below command, using latest version (time of writing that was: 36.4.1) of Sitespeed.
Please note that this has been happening for many versions but only now been able to create a good error report. We get the same error both locally and using GitHub Runners.

sitespeed.io --maxLoadTime 30000 -b firefox --firefox.includeResponseBodies all --firefox.preference privacy.trackingprotection.enabled:false --firefox.preference privacy.donottrackheader.enabled:false --firefox.preference browser.safebrowsing.malware.enabled:false --firefox.preference browser.safebrowsing.phishing.enabled:false --firefox.preference security.tls.version.min:1 --firefox.preference security.tls.version.max:1 --pageCompleteCheckNetworkIdle true --browsertime.screenshot false --screenshot false --screenshotLCP false --browsertime.screenshotLCP false --videoParams.createFilmstrip false --visualMetrics false --visualMetricsPerceptual false --visualMetricsContentful false --browsertime.headless true --silent true --utc true -n 1 https://webperf.se
```

### Log output

```shell
[2025-02-19 16:09:08] ERROR: Failed waiting on page https://gotene.se/?webperf-core=TLSv1.1 to finished loading, timed out after 30000 ms Error: Running page complete check 
return (function(waitTime) {
    try { 
            var end = window.performance.timing.loadEventEnd;
            var start= window.performance.timing.navigationStart;
            return (end > 0) && (performance.now() > end - start + waitTime);
    } 
    catch(e) {
        return true;
    }
[...]

PR #4454, opened 2025-02-27
Ensure we handle no response from website better

7h3Rabbit:
This a try to remedy the issue #4453.
It is probably not complete and not perfect but maybe something you can build from or tell me what to change to make it usable :)

Just trying to help and learn :)


soulgalore (commented):
Hi @7h3Rabbit thank you for the PR! Could you remove the changes in linux.yml so that only the fix for the breaking responses are there or are the test in the linux.yml an example where it fails?

7h3Rabbit:
@soulgalore 

>  Could you remove the changes in linux.yml
Yes

> are the test in the linux.yml an example where it fails?

Yes, I used it to validate the change and as you can see there are still one issue left (test is not marked as failed but if you expand it you can see that the code still crashes (but on imported packages so they probably need updating after this PR)

7h3Rabbit:
@soulgalore I have now removed the change in the linux.yml file

soulgalore:
Thank you @7h3Rabbit I will merge this later this week and release it asap. I've been busy moving the influxdb plugin to another repo this weekend so haven't had time yet with this PR, sorry, and thank you again :)

soulgalore:
Hi again @7h3Rabbit I've made https://github.com/sitespeedio/sitespeed.io/pull/4457 at least when I tried it fixes the problem for me. I changed so that if we do not have any data we do not send it at all. I'm thinking that makes more sense, hope that's ok for you.

7h3Rabbit:
@soulgalore Works great. I can't wait for next release with that PR included :D
I'm closing this PR.

PR #4499, opened 2025-04-15
Fixed issue with failing sanitizing

7h3Rabbit:
This PR is intended to solve that waterfall tab in page summary sometimes crashes when using:
`--browsertime.chrome.includeResponseBodies all`

When waterfall tab breaks, it also breaks all tabs on the right of it making it impossible to view plugin reports.

This PR changes the blacklisting approach of sanitizing browsertime.har file to instead removing/replacing problematic text property entirely. 

You can reproduce this error (before and after PR) by calling sitespeed.io with following command:
`node bin\sitespeed.js -b edge -n 1 --browsertime.chrome.includeResponseBodies all https://bollnas.se`

### DIFF
--- a/lib/plugins/html/templates/url/waterfall/includeHARinHTML.pug
+++ b/lib/plugins/html/templates/url/waterfall/includeHARinHTML.pug
@@ -1,10 +1,16 @@
 #output
 //- When we inline the data, we cut the HAR to just be one
 //- specific run (the median run) in summaryPageHAR.
-//- Inline script tags breaks the output
-- dahar = h.get(pageInfo.data, 'browsertime.run.har', summaryPageHAR)
-- dahar = JSON.stringify(dahar.log).split('</script>').join('&lt;/script&gt;')
-
+//- Inline script tags break the output
+- var dahar = h.get(pageInfo.data, 'browsertime.run.har', summaryPageHAR)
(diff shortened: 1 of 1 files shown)


7h3Rabbit:
@soulgalore Is there anything you want me to change in this PR?

soulgalore:
Hi @7h3Rabbit thank you for the PR! I want to have look tomorrow. I personally use the content in the HAR file a lot so I want to check if there's another way of doing it.

7h3Rabbit:
@soulgalore Great! 😄 Please note that the modification of `dahar` only affect the pug file. not the browsertime.har file.
I use the browsertime.har a lot as well 😄

soulgalore:
@7h3Rabbit could you see what makes it break?

7h3Rabbit:
@soulgalore it breaks the script tag containment, but I could not see exactly what it was.
After some search and pen testing it was clear to me that having a narrow blacklisting is not secure enough to use sitespeed.io to test other peoples websites as I was able to retrieve and run a lot of things when modifying the website called.

You use a different solution in an other pug file where you fetch the browsertime.har file from disk, if my PR is not good I would suggest looking into doing something similar here as well.

soulgalore:
The problem was an ending script tag. I made a PR https://github.com/sitespeedio/sitespeed.io/pull/4500 that fixes Bollnäs for me.

PR #4544, opened 2025-08-18
blog: Improve RSS feed

Krinkle:
* [x] Please describe your pull request

### Description

* Point rss discovery to canonical URL.
  Update discovery link to the current canonical URL for where the blog RSS feed lives. The redirect is kept for compatibility. This avoids feed readers from becoming unsure about the feed location by not re-crawling and re-fixing the redirect repeatedly.

* Fix 404 from https://www.sitespeed.io/feed/blog.xml to `https://www.sitespeed.io/feed.xml` which may confuse feed readers if they replace their link with the preferred self location, but then lose the subscription to an HTTP 404 error.

* Shorten the feed title and start with "Sitespeed.io" so that it sorts well alongside other blogs in a feed reader under the default suggested name.


soulgalore:
Thank you @Krinkle :)

PR #4581, opened 2026-01-14
docs: fix traffic_shaping link and add cyrilex regex resource

harsh9011:
This PR fixes a broken link to traffic_shaping.py and adds cyrilex as a regex testing resource in the documentation.

Related to #4272

### DIFF
--- a/docs/documentation/onlinetest/index.md
+++ b/docs/documentation/onlinetest/index.md
@@ -26,17 +26,17 @@ The online test is the simplest way to deploy your own version of sitespeed.io. 
 
 This method is ideal for small, medium, and large companies and organizations that also wants a GUI for running performance tests.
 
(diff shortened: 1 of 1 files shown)


soulgalore:
The https://www.cyrilex.com link do not work for me.

harsh9011:
Thanks for pointing that out — you’re right, the Cyrilex link seems to be unreachable.
I’ve updated the documentation and replaced it with regex101, which is a working regex testing tool.

Please let me know if this works for you now.

harsh9011:
i have replaced the broken Cyrilex link with regex101 and pushed the update last week
Please let me know if anything else is needed from my side

soulgalore:
So the traffic shaping file at Catchpoint is Polyform Shield 1.0.0 licensed but that was not the thing in WebPageTest so it is actually not the same :(  What about just removing the link?

PR #4583, opened 2026-01-15
feat(sustainable) : added countryCode option for country specific  CO₂ calculations

Harsh-Git39:
Makes users to calculate CO₂ emissions using a country-specific electricity grid mix via electricitymaps (through `@tgwf/co2`).

## Implementation

Added `sustainable.countryCode` as a plugin CLI option.
Passed the value into the CO₂ model configuration when creating the `co2` instance.
 Updated the sustainable plugin documentation with a implementation example .

## Testing and observations

I tested it with different country codes (IN, DE, NO). On very small pages (example.com) the difference is tiny due to rounding, but on larger pages like wikipedia.org the CO₂ values do change (for example, India ≈ 0.03730 vs Norway ≈ 0.03710), which shows the country-specific electricity mix is being used.
On lightweight pages (like example.com / wikipedia), the differences in CO₂ values are small, but they are consistently visible across runs, which suggests the country-specific electricity mix is being applied correctly. Heavier pages should make the differences more noticeable.

Thanks a lot for opportunity!!


soulgalore:
Hi @Harsh-Git39 ah cool, I'll have a look and test it this weekend!

PR #4759, opened 2026-05-23
Fix: Allow Unicode characters in aliases to prevent folder collisions (fixes #3880)

### LINKED ISSUE #3880: Can not use chinese charactor in alias, it will become '----' when creating the pages folder.
### Feature/improvement

The urls is too long and hard to read in HTML report, so I wish use alias, which is more easier to understood like native language - chinese for me. But I found when pass chinese charator to measure.start(url, '中文'), and multiple urls, it will come error like: index out of range when generating HTML report.

When looking into it, I found the error is because the page path is wrong with '----' instead of alias name. When I change alias to English name, it will be good.

Is there any way to use a more readable native charator in alias?

Thanks a lot for your reply.

UdeepChowdary:

- [ ] Update the documentation https://github.com/sitespeedio/sitespeed.io/tree/main/docs in another PR

### Description
Fixes #3880.

When passing non-ASCII characters (like Chinese or Cyrillic) to the `alias` parameter, the sanitization logic previously stripped them out and replaced them with dashes. This caused severe folder path collisions (e.g., multiple different Chinese aliases all compiling down to `----`), resulting in an `index out of range` error during HTML report generation.

### The Fix
* **Unicode Support:** Updated the replacement logic in `pathToFolder.js` to explicitly preserve Unicode letters (`\p{L}`) and numbers (`\p{N}`), while continuing to strip standard file-system unsafe characters.
* **Pathing Compliance:** Swapped native OS `path.join` for standard array `.join('/')` to enforce web-safe forward slashes across all environments.
* **Regression Tests Added:** Added test cases in `pathToFolderTests.js` to ensure Chinese, Cyrillic, and mixed ASCII/CJK aliases are preserved and produce distinct folder paths.

### DIFF
--- a/lib/core/resultsStorage/pathToFolder.js
+++ b/lib/core/resultsStorage/pathToFolder.js
@@ -85,8 +85,12 @@ export function pathToFolder(input, options, alias) {
   // pathSegments.push('data');
 
   for (const [i, seg] of pathSegments.entries()) {
-    if (seg) pathSegments[i] = seg.replaceAll(/[^\w.\u0621-\u064A-]/giu, '-');
+    // Keep Unicode letters (\p{L}), Unicode digits (\p{N}), dots, and hyphens.
+    // Everything else (filesystem-unsafe chars like <>:"/\|?* and bare spaces)
+    // is replaced with '-'. This fixes collisions when aliases use CJK or other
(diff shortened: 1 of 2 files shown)


soulgalore:
Thank you @UdeepChowdary

UdeepChowdary:
@soulgalore Thank you for the review and merging it! Happy to contribute more!

PR #4795, opened 2026-06-21
build: exclude npm-shrinkwrap.json from published package

### LINKED ISSUE #4592: npm-shrinkwrap.json causing issues with internal npm registry
### Have you read the documentation?

- [x] Yes, I've read the [how to make a reproducable bug guide](https://www.sitespeed.io/documentation/sitespeed.io/bug-report/)
- [x] Yes, I've read the [how to debug my script guide](https://www.sitespeed.io/documentation/sitespeed.io/scripting/#debug)

### URL

n/a

### What are you trying to accomplish

Hi,

The current version (39.4.2) of sitespeed.io (and many others as well) includes `npm-shrinkwrap.json`. It looks like this was introduced many years back, but it is preventing us from installing it in our company setup as we have an internal instance of the npm registry due to security issues.

I wish there was a better solution, something in the `.npmrc` file for example to disable it on our end, but there is unfortunately not. Until then having `npm-shrinkwrap.json` will prevent us from using `sitespeed.io`. Can you please consider removing the file?

Thank you!

### What browser did you use?

Other

### How to reproduce

```shell
On an internal npm registry do an install of the package
```

### Log output

```shell

```

fuleinist:
Fixes sitespeedio/sitespeed.io#4592

## Problem

The published \sitespeed.io\ npm tarball includes \
pm-shrinkwrap.json\. Some downstream users run internal npm proxies (e.g. corporate registries) that reject packages containing a shrinkwrap file, which blocks them from installing sitespeed.io.

## Fix

Keep \
pm-shrinkwrap.json\ in the repo (it's used by \
pm install\ for reproducible dev installs) but exclude it from the published package via \.npmignore\. This is the standard pattern for libraries that use shrinkwrap files.

## Verification

\
pm pack --dry-run\ on this branch:
- WITHOUT the fix: \
pm-shrinkwrap.json\ (388KB) is in the tarball contents.
- WITH the fix: \
pm-shrinkwrap.json\ is excluded; tarball size drops from ~984KB to 596.1KB; total files: 347.

\package.json\ does not define a \iles\ field, so \.npmignore\ is the sole authority for what ships. \
pm install\ inside the repo (post-clone) is unaffected because shrinkwrap is read from the working tree, not from the package contents.

## Diff

\\\diff
--- a/.npmignore
+++ b/.npmignore
@@ -9,4 +9,9 @@ release/*
 Gemfile
 Gemfile.lock
 wpr-record.log
-wpr-replay.log
\ No newline at end of file
+wpr-replay.log
+# Keep npm-shrinkwrap.json in the repo (used by \
pm install\ for
+# reproducible dev installs), but exclude it from the published package.
+# Some downstream users run internal npm proxies that reject packages
+# containing npm-shrinkwrap.json. See discussion in issue #4592.
+npm-shrinkwrap.json
\\\

Single file, 6 insertions, 1 deletion. No source code touched.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -10,3 +10,8 @@ Gemfile
 Gemfile.lock
 wpr-record.log
 wpr-replay.log
+# Keep npm-shrinkwrap.json in the repo (used by `npm install` for
+# reproducible dev installs), but exclude it from the published package.
+# Some downstream users run internal npm proxies that reject packages
+# containing npm-shrinkwrap.json. See discussion in issue #4592.
+npm-shrinkwrap.json


soulgalore:
Without the npm-shrinkwrap.json I don't see how we can make sure we can reproduce bugs, since we cannot guarantee that we use the same versions?

fuleinist:
Good point about reproducibility. The 
pm-shrinkwrap.json is still in the repo — this PR only excludes it from the published npm package (via \.npmignore\), so anyone cloning the repo still gets the pinned dependency tree for reproducing bugs. The package consumers get a clean install that respects their own lockfile.

If you'd prefer a different approach (e.g. keeping shrinkwrap in the published package), happy to adjust.

fuleinist:
Following up — any preference between the current approach (
pm-shrinkwrap.json stays in the repo via .npmignore exclusion in the published npm package) and keeping shrinkwrap in the published package itself?

If no strong preference, happy to close this PR in a week to free the slot. Otherwise, point me at the alternative and I'll update. Thanks for the time either way.

fuleinist:
Soft close — this PR has been waiting for maintainer feedback since Jun 24 with no further engagement from @soulgalore (2 follow-ups sent). Closing to free the slot. Happy to reopen if there's still interest. Thanks for the initial review.

fuleinist:
Closing as stated in the soft-close notice yesterday. No maintainer response received. Happy to reopen if there's interest.

PR #4896, opened 2026-09-16
Show CrUX ad metrics in the report and time-series export

### LINKED ISSUE #4895: Add CrUX Ads Metrics
### Feature/improvement

Google has a new CrUX API specifically for ads metrics. It would be useful to have this as an included feature alongisde the current CrUX web vitals metrics.

https://developer.chrome.com/docs/ads/metrics

soulgalore:
Google now reports ad count, ad density, ad CPU time and ad network weight through the CrUX API. They arrive in the same queryRecord response sitespeed.io already fetches, so users who collect CrUX data were getting them without seeing them.

The metrics only carry a p75 and have no good/poor thresholds, so the CrUX tab shows them as an ungraded tile row and the values are exported to Graphite and InfluxDB alongside the Web Vitals. Values are coerced to numbers because the API returns some as strings and the exporter drops non-numeric values.

Closes #4895

Co-authored-by: Claude Fable 5.1 noreply@anthropic.com


soulgalore:
Failing unrelated errors, lets fix that in other PRs.

